PR #315, opened 2014-01-07
Feature: Allow functions to be added to collection script domain

wuurrd:
Often one does repetitive things in the script domain, e.g. create a unique validator that one wishes to reuse, or a simple function that needs to be done on almost all collections, so this commit allows the users of deployd create generic functions they can inject into the script domain.

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -573,7 +573,7 @@ Collection.prototype.save = function (ctx, fn) {
 
 Collection.domainAdditions = {};
 
-Collection.prototype.addScriptDomainFunction = function(name, fn) {
+Collection.addScriptDomainFunction = function(name, fn) {
   Collection.domainAdditions[name] = fn;
 }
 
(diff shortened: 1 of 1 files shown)


ericfong:
Any update on this?

wuurrd:
@ericfong Sorry for the delay, hadn't checked my notifications for a while.

Added @luto 's fix, and made CI green.

andreialecu:
It looks like this can also be used for other things (objects, arrays, etc), not just functions. A possible suggestion for a better name could be `Collection.ExtendDomain()`. (I would also Pascal case it now that it is static)

lib/resources/collection/index.js
```diff
@@ -596,6 +602,7 @@ function createDomain(data, errors) {
     'this': data,
     data: data
   };
+  util._extend(domain, Collection.domainAdditions);
```
ericfong: shall we use underscore extend instead of non-public api from node util ?

andreialecu:
Closing in favor of https://github.com/deployd/deployd/pull/593

luto:
Thank you for taking care of this!

PR #384, opened 2014-07-29
fix(db): update mongodb version to fix crash when GETting large resultsets

jpgcc:
breaks sanitize behaviour (at least judging by the integration test results)


ericfong:
Shall we close this. Seems it duplicated with #467

NicolasRitouet:
indeed

Issue #370, opened 2014-06-07
Fail to start dpd 6.10 on Windows Server 2012 (64 bit)


rebornix:
I installed `dpd 6.10` through `npm` and I can create application successfully using CLI. 

However, when I ran `dpd` in cmd, I got `starting deployd v0.6.10...` and it just quited immediately. I didn't even see any error messages. 

I know we need mongodb to run deployd, mongodb has been installed already but I can find any configuration instruction in the guide. So I'm not sure whether I config it perfectly.

How can I fix or debug this?

Issue #364, opened 2014-05-26
dpd -p 3000 & does not start on ubuntu 12.04


ramstein74:
Hello, i installed deployd on ubuntu server 12.04 with no GUI , just ssh access ,
i create an app1
then 
dpd inside my app folder
dpd -p 3000 &
i only see 
"Starting deployd V0.6.10"

it stucks at this message!!!

if i press enter it terminates the process
[1] + Done    dpd -p 3000

How can i use deployd on ubuntu?

Issue #334, opened 2014-02-26
dpd fails to start on Windows 7


ctmay4:
I've read the other issues about users having trouble getting dpd to start.  However everyone seems to be on OS X or Linux.  I am running Windows 7 and do not have admin privileges.  I have node and mongodb already installed.  I installed dpd using npm:

> npm install deployd -g

Everything looks like it installed correctly:

> C:\Prj\test-dpd>dpd --version
> 0.6.10
> C:\Prj\test-dpd>mongo --version
> MongoDB shell version: 2.4.9

I start up dpd and get this:

> C:\Prj\test-dpd>dpd -d
> starting deployd v0.6.10...
> 
> C:\Prj\test-dpd>

There is no server responding to http://localhost:2403/dashboard.  A data directory is getting created in the project:

> C:\Prj\test-dpd\data>dir
> 02/26/2014  02:58 PM    <DIR>          .
> 02/26/2014  02:58 PM    <DIR>          ..
> 02/26/2014  02:49 PM        16,777,216 local.0
> 02/26/2014  02:49 PM         4,194,304 local.ns
> 02/26/2014  02:58 PM                 5 mongod.lock

I'm not sure where else to look.  I see people posting about starting with "DEBUG", but that method doesn't work on Windows.  I wish there was more feedback from the command-line to indicate the problem.  I'm not sure how to debug this issue.

ervo21:
i'm encountering a compatibility issue in installing deployd. Whenever i'm running 'dpd create sportsstore' it says that the version of my node.exe (in deployd installation folder) is not compatible in the version of my windows. I'm pretty sure that i've downloaded the correct version of node.exe..for deployd, i just selected the windows installer. How could i fix this? :(

ericfong:
Personally I have not tried in windows machine. 

Maybe you can also ask that in https://groups.google.com/forum/#!forum/deployd-users

ervo21:
Hi @ericfong , thank you!

Kvarkas:
or you can ask me? :) I am windows user - could you please specify steps that you are doing?

Kvarkas:
and try do update before start using ;) if got many errors - update json file from github

ervo21:
@Kvarkas thanks in advance. :)
 
In the same location of my nodejs installation folder i'm entered the 'dpd create sportsstore' and it says: 

"C:\Program Files\nodejs\sportsstore>dpd create sportsstore
[...]

Issue #262, opened 2013-09-02
Starting deployd


enriquemorenotent:
Im having trouble getting started with deployd.

I create the app, go into the directory and run `dpd`. First I got this message

> starting deployd v0.6.10...

And then the prompt returns.

When I go to http://localhost:2403 I only get an error saying "Oops! Google Chrome could not connect to localhost:2403"

Did I miss something?

PR #434, opened 2014-12-03
style(db): improve error messages when mongod is not available

ericfong:
Example:

```
Cannot connect to mongodb { host: 'localhost', port: 27017, name: 'db_name' }
```


ericfong:
I think 30e603c16ecb4342e8e1b01472eb6de304a994b3 fix most of the problem from #370 #364 #334 #262

PR #437, opened 2014-12-09
Feature: dpd script can change its url base

iranreyes:
# Modification

In my current project I am using Deployd as server of test (like a mock of the remote server). I communicate remote with the Deployd server through HTTP, but to have more control about the data I decided include the dpd.js. So, to include the script I must reference the script in my html like:

```HTML
<script src="http://localhost:2403/dpd.js"></script>
```

The problem if that if my server run on another port the dpd script try to recover data from the port of my app instead of the port where it run, in this case 2403. 

To Improve the script I add a function that let you configure the BaseUrl, the default behavior is keep it and you can change it calling this function at the begin of your application like this:

```
dpd.setupBaseUrl(protocol, hostname, port)
```

If you omit protocol will be used ```window.location.protocol```, if you omit hostname will be used ```window.location.hostname``` and port = ```window.location.port```

In my real case I change it like this:

```
dpd.setupBaseUrl(null, null, 2403)
```

### DIFF
--- a/clib/dpd.js
+++ b/clib/dpd.js
@@ -2,10 +2,9 @@
 
   if (!window._dpd) window._dpd = {};
 
-  var root = window.location.protocol + '//' + window.location.hostname;
-  if (window.location.port !== '') {
-    root += ':' + window.location.port;
-  }
+  var root = null;
+  
+  setupBaseUrl();
 
   var consoleLog = (typeof console !== 'undefined') && console.log;
 
@@ -13,6 +12,17 @@
   var socket = io.connect(root);
 
   var BASE_URL = '/';
+  
+  function setupBaseUrl(protocol, hostname, port){
+	  protocol = protocol || window.location.protocol;
+	  hostname = hostname || window.location.hostname;
+	  port = port || window.location.port;
+  
+	  root = protocol + '//' + hostname;
+	  if (port !== '') {
+		root += ':' + port;
+	  }
+  }
 
   function normalizeArray(parts, allowAboveRoot) {
     // if the path tries to go above the root, `up` ends up > 0
@@ -270,6 +280,8 @@
 
     return r;
   };
+  
+  window.dpd.setupBaseUrl = setupBaseUrl;
 
   window.dpd.on = function() {
     socket.on.apply(socket, arguments);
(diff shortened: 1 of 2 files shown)


andreialecu:
@iranreyes since you're using angularjs, have a look at https://github.com/slively/dpd-angular-cache. I had the same requirement as you and sent a PR to the maintainer to include this functionality (and more).

ericfong:
Actually, I also need this feature for my projects.

I would like to base on your PR or submit another PR for this

andreialecu:
To solve the problem about connecting to socket.io multiple times, the connection could be delayed until the first call to `dpd.on` `dpd.off` or `dpd.socketReady`. That way you can call `dpd.setupBaseUrl() `before setting up the handlers later on in the code and there should be no issue

andreialecu:
Here's a gist of the modifications I was talking about. All tests pass.

https://gist.github.com/andreialecu/45eba279e4ccc009a5af

ericfong:
Good gist.

Shall we have:
```
dpd.setBaseUrl(path) {}
dpd.getBaseUrl(path) {}
```
instead of
```
function setupBaseUrl(protocol, hostname, port) {}
```
?

andreialecu:
I think it's better if the parts of the URL are separated, so you can just override the protocol, or port, but keep the hostname. Say you have your website at http://mydomain.com and the api at https://api.mydomain.com:1337.

andreialecu:
I updated the gist, relevant code:

```javascript
  function setupBaseUrl(options) {
    options = options || {};
    if (typeof options === "string") {
      root = options;
      return;
    }
    
    var protocol = options.protocol || window.location.protocol;
    var hostname = options.hostname || window.location.hostname;
    var port = options.port || window.location.port;
    
    root = protocol + '//' + hostname;
    if (port !== '') {
      root += ':' + port;
    }
  }
```

ericfong:
Let's continue in #490

NicolasRitouet:
#490 is the followup of this features, among others.
I'll close this PR.
Thanks @iranreyes

clib/dpd.js
```diff
@@ -13,6 +12,17 @@
   var socket = io.connect(root);
```
jcguarinpenaranda: dpd.init(...)  is not working :S now, and dpd.setupBaseUrl(..) neither

Issue #453, opened 2014-12-18
Race Condition/Bug in getConnection


andreialecu:
I have an app that sends a lot of requests immediately on load to dpd, and this causes dpd to bail out with this error more often than not:

```
GET <snip> Error: db object already connecting, open cannot be called multiple times
    at Db.open (<snip>\node_modules\deployd\node_modules\mongodb\lib\mongodb\db.js:234:11)
    at getConnection (<snip>\node_modules\deployd\lib\db.js:123:13)
    at collection (<snip>\node_modules\deployd\lib\db.js:156:3)
    at SessionStore.Store.insert (<snip>\node_modules\deployd\lib\db.js:266:3)
    at SessionStore.createSession (<snip>\node_modules\deployd\lib\session.js:86:10)
    at <snip>\node_modules\deployd\lib\server.js:101:21
    at <snip>\node_modules\deployd\lib\util\http.js:64:9
    at <snip>\node_modules\deployd\node_modules\corser\lib\corser.js:219:21
    at exports.setup (<snip>\node_modules\deployd\lib\util\http.js:24:3)
    at Server.handleRequest (<snip>\node_modules\deployd\lib\server.js:98:3)
```

Upon investigating, I noticed that there is a problem here:
https://github.com/deployd/deployd/blob/master/lib/db.js#L122

When multiple calls go into getConnection, and the call goes through to db._mdb.authenticate() below, db.connected is set to true in an async callback, which sometimes is not fired until after another call has already entered getConnection and sees db.connecting as false, but db.connected as still yet undefined.

Does it make sense?

Anyway, the fix is to change the function to the following:

``` javascript
function getConnection(db, fn) {
  if(db.connected) {
    fn(null, db._mdb);
  } else if(db.connecting) {
    db.once('connection attempted', function (err) {
      fn(err, db._mdb);
    });
  } else {
    db.connecting = true;
    db._mdb.open(function (err) {
      db.emit('connection attempted', err);
      if(err) {
        db.connected = false;
        throw err;
      } else {
        // check for credentials
        var credentials = db.options.credentials;
        if (credentials && credentials.username && credentials.password) {
          db._mdb.authenticate(credentials.username, credentials.password, function (err) {
            if (err) {
              db.connected = false;
              throw err;
            }
            db.connecting = false;
            db.connected = true;
            fn(null, db._mdb);
          });
        } else {
          db.connecting = false;
          db.connected = true;
          fn(null, db._mdb);
        }
      }
    });
  }
}
```

PR #454, opened 2014-12-18
Fix a "race condition" resulting in bugs when opening a connection under heavy load

andreialecu:
See https://github.com/deployd/deployd/issues/453


spinningarrow:
I can't figure out how this works; doing
   dpd.users.on('login', function () { debugger; });
does nothing for me.

I'm running version `0.8.0` on OSX. Any ideas?

PR #455, opened 2014-12-20
test: fix the relative path

ericfong:


NicolasRitouet:
Landed as b70267424e83666f9a7a4383fe5036cbe1074aea
Thanks @ericfong

PR #464, opened 2015-01-02
Updated ace to "package 02.28.2014"

simpleTechs:
Also moved all the ace-related stuff into it's own subdir (for future updates)
Fixed a bug where the new version of ace would not show any text at all

This updates restores compatibility with most modern browsers and fixes a bug in safari on OS X where the fontsize shown would differ from the fontsize ace uses internally.


NicolasRitouet:
Landed as 4f3fe132dd8b8eeb1b75a9e536ccd486dd71c1d1

@simpleTechs Thanks for your contribution :+1:

PR #465, opened 2015-01-02
correctly delete cookies, update cookies-dependency

simpleTechs:
fixed a bug that would cause a logout call to be handled incorrectly. instead of removing the session-cookie altogether, only another cookie (with the same name `sid`) was set leading to undefined behavior.

This requires an updated version of the cookies-library, as overwrite can only be specified in 0.4.x and defaulted to `false` before.


NicolasRitouet:
Landed as ebad9fa552d067c2695dfcef62431c951bf590e1

@simpleTechs 

PR #467, opened 2015-01-03
Update mongodb to 2.0.9. Rewrite database code to use MongoClient.

andreialecu:
Rewrote database connection code/queue to use promises in order to fix a race condition (issue #453)


ericfong:
Great Thx. 

Tested (both unit test and integration test)
mac 10.10.1
node v0.10.35
mongodb 2.6.6
All passed.

Question:
version of mongodb driver on npm repo is 1.4.26
on github is 2.0.x
Do you know what is their status now?

lib/db.js
```diff
@@ -57,7 +58,7 @@ db.create = function (options) {
 
 function Db(options) {
   this.options = options;
-  this._mdb = new mongodb.Db(this.options.name, new mongodb.Server(this.options.host, this.options.port));
+  this.connectionString = this.options.connectionString || "mongodb://" + (this.options.credentials ? this.options.credentials.username + ":" + this.options.credentials.password + "@" : "") + this.options.host + ":" + this.options.port + "/" + this.options.name;
```
ericfong: I think it is nicer to have connectionString then a _mdb object.
But this line may be too long.

andreialecu:
@ericfong I refactored that line and pushed again, check it out now. https://github.com/deployd/deployd/pull/467/files#diff-be228df81b99a384aa50ad50bde7fca0R62

NicolasRitouet:
Landed as 0f6c1123a36f164c30ca047a1ece374f9245a6dd

@andreialecu 

ericfong:
Really nice PR. Thanks!

n321203:
Some info that might help someone: 

I am running on Heroku and the Composer/MongoHQ add-on. After this upgrade, I kept getting Mongo authentication errors and couldn't connect to my database. The reason was that my database password contained "@", which MongoClient seem to encode to "%40" (I don't know how it was treated previously)

After changing password (removing the @) everything works fine again. 
Lesson learned: Use a hard-to-guess password, but keep it within the usual characters...

andreialecu:
@n321203 nice catch! The password actually gets (properly) encoded here: https://github.com/deployd/deployd/pull/467/files#diff-be228df81b99a384aa50ad50bde7fca0R64 by the url.format() call.

And, actually this can be fixed by adding: uri_decode_auth: true to the second parameter of MongoClient. I'll have a PR to fix this shortly.

n321203:
Great – I was running the master branch of deployd but not the the mongodb client – that might explain it. I found the bug by inspection the username & password right before the DbCommand.createAuthenticationCommand in mongodb_cr.js, and saw that the @ was encoded.

PR #471, opened 2015-01-03
Use node crypto to generate session identifier.

andreialecu:


lib/session.js
```diff
@@ -44,8 +44,8 @@ function SessionStore(namespace, db, sockets) {
 util.inherits(SessionStore, Store);
 exports.SessionStore = SessionStore;
 
-SessionStore.prototype.createUniqueIdentifier = function() {
-  return uuid.create(128);
+SessionStore.prototype.createUniqueIdentifier = function () {
+  return crypto.randomBytes(64).toString('hex');
```
ericfong: 64 bytes may be too long. 16 - 20 bytes is already enough?

andreialecu: 64 bytes means a string of length 128 here, which is the same as it was before. I didn't want to change it from what it was.

ericfong:
Landed as 27340fa19df26899deb2400ec50072fd9e1d9cda
Thanks for your contribution!

This commit keeps deployd DRY

PR #472, opened 2015-01-03
Chore: Update rewire to 2.1.3

NicolasRitouet:
@ericfong 
all tests (unit and integration) pass with these changes, do you see any problem with the deletion of the setters I made?


ericfong:
Landed as c43f0dd368efc297fa794f994e89b3fabd3b49d3
Thanks for your contribution!

Issue #98, opened 2012-10-25
Deployd won't start under MacOS X ML 10.8.1


dseredinov:
Hey guys! 

As mentioned in twitter (@DmitrySeredinov), here is a console log:
```
[dals@Dals-Mac-mini deployd-projects]$ DEBUG=* dpd create foo-world -d
starting deployd v0.6.7...
  mongod starting mongod +0ms
  mongod all output going to: /usr/local/var/log/mongodb/mongo.log
 +857ms
```
That's all. Nothing more is happens.
Tail of mongod log below:
```
[dals@Dals-Mac-mini deployd-projects]$ tail -F /usr/local/var/log/mongodb/mongo.log
Thu Oct 25 11:57:26 [initandlisten] MongoDB starting : pid=73136 port=4426 dbpath=./data 64-bit host=Dals-Mac-mini.local
Thu Oct 25 11:57:26 [initandlisten] 
Thu Oct 25 11:57:26 [initandlisten] ** WARNING: soft rlimits too low. Number of files is 256, should be at least 1000
Thu Oct 25 11:57:26 [initandlisten] db version v2.2.0, pdfile version 4.5
Thu Oct 25 11:57:26 [initandlisten] git version: f5e83eae9cfbec7fb7a071321928f00d1b0c5207
Thu Oct 25 11:57:26 [initandlisten] build info: Darwin bs-osx-106-x86-64-1.local 10.8.0 Darwin Kernel Version 10.8.0: Tue Jun  7 16:33:36 PDT 2011; root:xnu-1504.15.3~1/RELEASE_I386 i386 BOOST_LIB_VERSION=1_49
Thu Oct 25 11:57:26 [initandlisten] options: { bind_ip: "127.0.0.1", config: "/usr/local/etc/mongod.conf", dbpath: "./data", logappend: "true", logpath: "/usr/local/var/log/mongodb/mongo.log", nojournal: true, pidfilepath: "./.dpd/pids/mongod", port: 4426 }
Thu Oct 25 11:57:26 [initandlisten] Unable to check for journal files due to: boost::filesystem::basic_directory_iterator constructor: No such file or directory: "./data/journal"
Thu Oct 25 11:57:26 [websvr] admin web console waiting for connections on port 5426
Thu Oct 25 11:57:26 [initandlisten] waiting for connections on port 4426
```

System information:

- OS: MacOS X Mountain Lion 10.8.1, build 12B19 on Intel Core i5 CPU, 8 GB RAM.

[...]

PR #473, opened 2015-01-04
docs(dpd-cli): docs for what to check when mongod is missing

ericfong:
Try to solve #98 by telling new users what they should check.

### DIFF
--- a/bin/dpd
+++ b/bin/dpd
@@ -158,7 +158,7 @@ function start(file) {
 
   function startup (err) {
       if (err) {
-        console.log("Failed to start MongoDB");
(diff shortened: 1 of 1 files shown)


bin/dpd
```diff
@@ -158,7 +158,7 @@ function start(file) {
 
   function startup (err) {
       if (err) {
-        console.log("Failed to start MongoDB");
+        console.log("Failed to start MongoDB (Make sure your -m OR --mongod path is right OR env $PATH points to the mongod executable)");
```
NicolasRitouet: If I don't know what's -m, it doesn't help.
I would explain that -m is an argument when you start deployd.
Or even better: just display a link to a dedicated page on our documentation.

PR #476, opened 2015-01-04
feature(tests): run integration tests when running npm test

andreialecu:
Adds --execute parameter to dpd CLI which runs a specified command and then exits.
Update package.json to run integration tests automatically as part of `npm test` using new -e parameter of CLI.

This will also make travis run the integration tests as part of a build.

### DIFF
--- a/bin/dpd
+++ b/bin/dpd
@@ -37,7 +37,7 @@ program
   .option('-P, --mongoPort [mongoPort]', 'mongodb port to connect to')
   .option('-n, --dbname [dbname]', 'name of the mongo database')
   .option('-a, --auth', 'prompts for mongo server credentials')
-  .option('-t, --test [testCmd]', 'run a specified command and exit')
+  .option('-e, --execute [cmd]', 'executes a command after starting and exit')
   ;
 
 /**
@@ -79,7 +79,7 @@ function start(file) {
     , env = program.environment || process.env.DPD_ENV || 'development'
     , retries = 0
     , credentials
-    , testCmd = program.test
+    , executeCommand = program.execute
     ;
 
   if (!port) {
@@ -178,10 +178,10 @@ function start(file) {
 
       function onListening () {
         console.info('listening on port', options.port);
-        if (program.test) {
-          console.log("Running command: " + program.test);
-          sh.exec(program.test, function (code, output){
-		    console.log("Command exited with code: " + code);
+        if (executeCommand) {
+          console.log("Running command: " + executeCommand);
+          sh.exec(executeCommand, function (code, output){
+            console.log("Command exited with code: " + code);
             stop(code); 
           });
         } else {
--- a/bin/dpd
+++ b/bin/dpd
@@ -181,7 +181,8 @@ function start(file) {
         if (program.test) {
           console.log("Running command: " + program.test);
           sh.exec(program.test, function (code, output){
-            process.exit(code); 
+		    console.log("Command exited with code: " + code);
+            stop(code); 
           });
         } else {
           var commands = repl(dpd);
--- a/bin/dpd
+++ b/bin/dpd
@@ -18,7 +18,8 @@ var program = require('commander')
   , open = require('../lib/util/open')
   , semver = require('semver')
   , Server = require('../lib/server')
-  , upgrade = require('doh').upgrade;
+  , upgrade = require('doh').upgrade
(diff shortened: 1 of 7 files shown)


andreialecu:
Interestingly there seems to have been some sort of fluke in running the integration tests for this build: https://travis-ci.org/deployd/deployd/builds/45886173

I did a minor spacing fix and amended the commit and repushed, the initial build went fine, but this second one failed with an error in an integration test. 

There seem to be random failures because of the async package being upgraded to 0.9.0. I have noticed them before in my testing as well. Something is off with the async package.

andreialecu:
Also, the Error loading resource errors from mocha-phantomjs cannot be turned off currently, but I have submitted a PR to mocha-phantomjs to fix it: https://github.com/metaskills/mocha-phantomjs/pull/169

I have a commit at https://github.com/andreialecu/mocha-phantomjs/commit/41e6dd725272724cef31982977cf4fdb9696dc74 which adds the -i switch to hide those errors and make the log a lot cleaner. It could be put straight into package.json as a github reference until the PR makes it into mocha-phantomjs proper.

andreialecu:
Another minor unrelated change and all tests passed: https://travis-ci.org/deployd/deployd/builds/45888892

andreialecu:
Interesting that this build failed: https://travis-ci.org/deployd/deployd/builds/46467404 , because this has been a bug that has plagued my tests randomly and I think I tracked it down. 

It's unrelated to any of these changes, but I think it is a race condition that starts here:
https://github.com/deployd/deployd/blob/master/lib/session.js#L85

Session is inserted, but before it actually gets a chance to get committed to the database another request may already enter through here: https://github.com/deployd/deployd/blob/master/lib/session.js#L67 and then call this: https://github.com/deployd/deployd/blob/master/lib/session.js#L70 while then ending up here:
https://github.com/deployd/deployd/blob/master/lib/session.js#L74 where s is still null and gets saved into the session that way resulting into a crash later on when accessing it here: https://github.com/deployd/deployd/blob/master/lib/session.js#L257 because data would be null.

I'll submit another PR for the fix.

NicolasRitouet:
The goal here is to run the integration tests automatically along with the unit tests.
For that, we need to be able to start and stop Deployd server within our CI workflow.
@andreialecu's solution is to add a param to the dpd CLI that will run any script after starting the app.
I would love not to modify the CLI api to achieve this.

[Supertest](https://github.com/tj/supertest) for example, can take an express app as param, start and stop it to run the tests on it.
I'm sure we can find a good solution for this.

andreialecu:
The problem is not starting deployd, but mongo along with it. The benefit of thIs implementation is that it also verifies that the CLI is working properly and it is the DRYest way of doing it.

NicolasRitouet:
Starting mongo manually is not an issue, travis CI will do it for us, and if somebody is skilled enough to run the tests and prepare a PR for us, he should have a mongodb running and easy to start.

andreialecu:
Travis has nothing to do with this PR. It's only for devs to be able to locally run npm test and run the integration tests on an isolated mongo process, like the one the CLI is already designed to start.

The fact it also works on Travis is just a side effect.

NicolasRitouet:
Having the integration tests on travis CI was the best advantage of this PR imho.

andreialecu:
By side effect, that  works indeed. But personally it doesn't make my life as olp working on

As a person who has contributed dozens of tests lately and changed some core functionality, the local npm test integration is invaluable to me. 

Would have saved me hours in my workflow being able to verify changes immediately on both test realms rather than having to keep starting and killing dpd every single time I needed to test something.

NicolasRitouet:
I'm also running all tests (unit and integration) many times a day and would save some time with this feature, but I don't want to take a rash decision and change our API for that.
Anyway, let's not pollute this thread, I'd like to discuss here about other alternatives, we can discuss about the rest on the chat.
If we list all alternatives and find nothing better, we can reconsider adding a param to the CLI.

andreialecu:
Updated to not change the CLI.

ericfong:
Landed as 99aa2b514a2391a2aebbb4964ce4ed978f4f2596

Run ```npm test``` around 20 times. All passed. No zombie process leftover. 
~23s - 30s for each npm test in my Mac

Thanks for your contribution!

NicolasRitouet:
Yeah :)

Thanks Andrei !

PR #477, opened 2015-01-05
refactor(config-loader): enhance the 'resources' dir reading

ericfong:


lib/attach.js
```diff
@@ -123,9 +123,12 @@ function attach(httpServer, options) {
       });
     });
   };
+  
+  
+  var serverpath = server.options.server_dir || fs.realpathSync('./');
 
   server.route = function route(req, res) {
-    config.loadConfig('./', server, function(err, resourcesInstances) {
+    config.loadConfig(serverpath, server, function(err, resourcesInstances) {
```
NicolasRitouet: Before this change, if you set server.options.server_dir to something different than ```.```, you got an error, right?
Can we have a test to cover that?

PR #478, opened 2015-01-06
Feature/Fix: Allows changes to user data from UserCollection Login event.

andreialecu:
Refactor login code into separate method in user-collection.js.
Improve integration tests for Login event.

This PR is important for the Login event, without it most of the actual functionality you could use it for is impossible. See the included tests for examples on how it can be used.

### DIFF
--- a/lib/resources/user-collection.js
+++ b/lib/resources/user-collection.js
@@ -169,29 +169,13 @@ function handleLogin(uc, ctx) {
       }
     }
 
-    if (user) {
-      // a user with this username exists
-      delete userClone.password;
-      if (uc.checkHash(uc, user, credentials) === true) {
-        domain.success = true;
-        delete user.password; // make sure the password is not included in any sort of response
-        
-        function loginDone(err) {
-          if (err) return ctx.done(err);
-          saveUser(function (err) {
-            if (err) return ctx.done(err);
-            debug('logged in as %s', credentials.username);
-            ctx.session.set({ path: path, uid: user.id }).save(ctx.done);
-          });
-        }
-
-        if (uc.events.Login) {
-          uc.events.Login.run(ctx, domain, loginDone);
-        } else {
-          loginDone();
-        }
-        return;
-      }
+    function loginDone(err) {
+      if (err) return ctx.done(err);
+      saveUser(function (err) {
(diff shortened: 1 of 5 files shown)


lib/resources/user-collection.js
```diff
@@ -166,6 +138,78 @@ UserCollection.prototype.handle = function (ctx) {
   }
 };
 
+function handleLogin(uc, ctx) {
```
ericfong: I suggest 
```
UserCollection.prototype.handleLogin = function(){}
```
so people like me can override it

lib/resources/user-collection.js
```diff
@@ -166,6 +138,78 @@ UserCollection.prototype.handle = function (ctx) {
...
+  debug('trying to login as %s', credentials.username);
+  
+  uc.store.first({ username: credentials.username }, function (err, user) {
+    if (err) return ctx.done(err);
+    // keep a clone of the user so we can compare it later to see if any changes were made in the login event
+    var userClone = user ? _.clone(user) : null
+      , domain = { 'me': userClone, 'data': userClone, 'success': false };
+    
+    // checks if the user was changed in the login event and saves it if it was
+    function saveUser(fn) {
```
ericfong: shall we call this as ```checkAndSaveUser```?

And move loginDone, loginFail and saveUser to another function 
```UserCollection.prototype.afterLoginEvent = function(){}```

Reason:

In my case, I call 
```
uc.store.first({$or:[{username:credentials.username}, {tel:credentials.username}, {email:credentials.username}]}, 
```
in order to login by email OR tel OR username. 
And also handle forget password and robot access in here. 

I would like to do that by overriding ```handleLogin```, and then, I can call  the Login event and afterLoginEvent function.

andreialecu: I'm not sure moving loginDone, loginFail and saveUser is right for this purpose. 

If I just put uc.store.first({ username: credentials.username }) in a prototype, then you can override that prototype and chance it for your needs in one place, and it's simpler.

Do you have other requirements this wouldn't satisfy?

ericfong:
Landed as 616b176fe3526ddbb1a94bcc577b467eded2ef37

PR #479, opened 2015-01-06
Fix: Allow changed() function to work properly with collection properties of type object

andreialecu:
changed('prop') in event scripts would return true even if the object was not changed. This fixes it by using underscore's deep object comparison functionality.

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -448,7 +448,7 @@ Collection.prototype.save = function (ctx, fn) {
 
   domain.changed =  function (property) {
     if(domain.data.hasOwnProperty(property)) {
-      if(domain.previous && domain.previous[property] === domain.data[property]) {
+      if(domain.previous && _.isEqual(domain.previous[property], domain.data[property])) {
         return false;
       }
 
--- a/test-app/public/test/collection.test.js
+++ b/test-app/public/test/collection.test.js
@@ -1090,6 +1090,25 @@ describe('Collection', function() {
       });
     });
 
+    it('should not return true when a value of type object has not changed', function(done) {
+      dpd.changed.post({name: 'irrelevant', data: { 'key': '$NO_CHANGE'} }, function (c) {
+        dpd.changed.put(c.id, { data: { 'key': '$NO_CHANGE'} }, function (c) {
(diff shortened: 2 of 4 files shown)


lib/resources/collection/index.js
```diff
@@ -448,7 +448,7 @@ Collection.prototype.save = function (ctx, fn) {
 
   domain.changed =  function (property) {
```
ericfong: Shall we 
```js
                    var args = Array.prototype.slice.call(arguments);
                    for (var i = 0, ii = args.length;  i < ii; i++) {

// The original code...

                    }
                    return false;
```

So that we can call ```changed('fieldA', 'fieldB')``` to detect ```fieldA``` OR ```fieldB``` is changed

Why OR?
In one-to-many or many-to-many relationship, actions are more likely to be triggered by sub-set of fields on a item.

andreialecu: That would be nice I guess, it's a feature that doesn't break anything. I can update the PR with it if you want.

ericfong: I go for OR case. 

I would like to know anyone will go for AND case and reason for that.

andreialecu: I still think it's somewhat unintuitive and can lead to issues.

How about changedAny('fieldA', 'fieldB') and changedAll('fieldA', 'fieldB') as two separate methods? It would be just as simple to implement.

I can think about things like custom logic for notifying an admin if ALL the specified fields are changed, but I can definitely see the ANY case being just as useful.

As an example, I could have:
```
if (changedAll('availabilityStatus', 'availabilityUntil')) {
  ..logic
} else {
  cancel('both availabilityStatus and availabilityUntil to be updated simultaneously');
}
```

ericfong: changedAny & changedAll can be in another PR. Then, we can keep this PR unchanged.

ericfong:
In case anyone consider performance, 
according to https://github.com/jashkenas/underscore/blob/master/underscore.js#L1083
Should not have bit performance different

ericfong:
landed as 46f518dc876e723c6c87c70a87b6ef7278b6dab4
Thanks for your contribution!

PR #481, opened 2015-01-09
fix(session): Refactor session creation code to use promises in order to fix a possible race condition while inserting a session to the database.

andreialecu:
Upon auditing the code I found a potential issue which also caused this build to fail: https://travis-ci.org/deployd/deployd/builds/46467404

The explanation for the change is here: 
https://github.com/deployd/deployd/pull/476#issuecomment-69373781

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -3,8 +3,8 @@ var Store = require('./db').Store
 ,	uuid = require('./util/uuid')
 , Cookies = require('cookies')
 , EventEmitter = require('events').EventEmitter
-, debug = require('debug')('session');
-
+, debug = require('debug')('session')
+, Q = require("q");
 /*!
 * A simple index for storing sesssions in memory.
 */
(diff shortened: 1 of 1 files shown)


lib/session.js
```diff
@@ -3,9 +3,8 @@ var Store = require('./db').Store
 , Cookies = require('cookies')
 , EventEmitter = require('events').EventEmitter
 , crypto = require('crypto')
-, debug = require('debug')('session');
-
-/*!
+, debug = require('debug')('session')
+, Q = require("q");/*!
```
ericfong: use bluebird-q  like #484 ?

NicolasRitouet: Can you fix this comment?

ericfong:
landed as 3080de2faaeee6dc2d22e6e8f201b60cbf26bff2
Thanks for your contribution!

PR #483, opened 2015-01-10
Fix(config-loader): Memory leak and inefficiency

andreialecu:
Every second HTTP request would reload the entire resources dir resulting in potential memory leaks.

### DIFF
--- a/lib/config-loader.js
+++ b/lib/config-loader.js
@@ -20,8 +20,6 @@ var fs = require('fs')
 module.exports.loadConfig = function(basepath, server, fn) {
   var resources = server.__resourceCache || [];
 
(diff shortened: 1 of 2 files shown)


andreialecu:
Note that resources are still being reloaded in case of a configuration change via dashboard from here: 
https://github.com/deployd/deployd/blob/master/lib/resources/internal-resources.js#L88

NicolasRitouet:
Looks good!
Thanks.

NicolasRitouet:
Landed as 6e7654e5adcb80c47764ce3d48dfcecd1563b207

Thanks @andreialecu :+1:

PR #484, opened 2015-01-11
feature(internal-client): Promise support for internal dpd client.

andreialecu:
Switched from Q to Bluebird for the promise library.

### DIFF
--- a/lib/script.js
+++ b/lib/script.js
@@ -159,7 +159,8 @@ function wrapAsyncFunctions(asyncFunctions, sandbox, events, done, sandboxRoot) 
         var callback;
         var callbackIndex;
         var result;
-        
+        var callbackAdded = false;
+          
         for(var i = 0; i < args.length; i++) {
           if(typeof args[i] == 'function') {
             callback = args[i];
@@ -167,9 +168,10 @@ function wrapAsyncFunctions(asyncFunctions, sandbox, events, done, sandboxRoot) 
             break;
           }
         }
-
+        
         if (typeof callback === 'function') {
           events.emit('addCallback');
+          callbackAdded = true;
           args[callbackIndex] = function() {
             if (sandboxRoot._error) return;
             try {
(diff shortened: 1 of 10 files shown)


lib/script.js
```diff
@@ -83,7 +83,7 @@ Script.prototype.run = function (ctx, domain, fn) {
       callbackCount++;
     });
 
-    events.on('finishCallback', function() {
+    events.on('finishCallback', function () {
       callbackCount--;
```
ericfong: if the wrap function calls without Promise or callback,
the ```callbackCount``` may be wrong, 
as the number of call to ```addCallback``` and ```finishCallback``` is not in pair?

lib/script.js
```diff
@@ -182,7 +182,12 @@ function wrapAsyncFunctions(asyncFunctions, sandbox, events, done, sandboxRoot)
...
         } else {
-          args.push(function() {
+          // CONVENTION:
+          // functions that use promises start with the name 'promisedXXXX', add a callback counter for them
+          if ((/^promised/).test(asyncFunctions[k].name)) {
+            events.emit('addCallback');
+          }
+          args.push(function () {
             if (sandboxRoot._error) return;
             events.emit('finishCallback');
```
ericfong: if the wrap function calls without Promise or callback,
the ```callbackCount``` may be wrong, 
as the number of call to ```addCallback``` and ```finishCallback``` is not in pair?

NicolasRitouet:
Can you explain why you needed to switch from Q to bluebird?

andreialecu:
@NicolasRitouet bluebird has this method: 
https://github.com/petkaantonov/bluebird/blob/master/API.md#binddynamic-thisarg---promise

It is needed to allow passing the `thisArg` into the thenables so that you can do:
https://github.com/deployd/deployd/pull/484/files#diff-6f7632d950e9c2863efdb29ce29b1a24R2

This is where it was needed:
https://github.com/deployd/deployd/pull/484/files#diff-7e23141f9dd7c0c72d781aa717bc1a7eR138

lib/script.js
andreialecu: Yes, this was like this from before, I didn't want to touch it because I don't know what the purpose is, maybe there is some quirk where it is needed. It seems on purpose but it may very have been a bug:

In deployd previous to this PR, If you called the internal client **without** ( eg `dpd.collection.get(123, null)` ) a callback, the callback count would indeed be wrong, it would keep substracting 1 on each call. 

However, because of line 187-188 in this PR, this is now fixed because we increment callbackCount anyway for *promised* methods. So a function returning promises will always increment the callback count, and when the promise is resolved, it will call the supplied callback to decrement the callback count.

andreialecu: We hook onto the promise to call this: https://github.com/deployd/deployd/pull/484/files#diff-7e23141f9dd7c0c72d781aa717bc1a7eR90 when the promise is finished, which guarantees that the callback count is decremented.

andreialecu:
I will refactor the code in a bit, I got a better idea on how to check if a promise is returned.

NicolasRitouet:
Would it be complicated to migrate completely to bluebird without using bluebird-q?

andreialecu:
I will close this one because I've been working with @simpleTechs on a better solution at #494.

PR #485, opened 2015-01-13
Fix(user-collection): Do not call other events (put/validate) when a user is changed from login event.

andreialecu:


lib/resources/user-collection.js
```diff
@@ -164,6 +164,9 @@ UserCollection.prototype.handleLogin = function (ctx) {
         // create a new context and set the body to our user so that we can call save on the collection
         var newCtx = _.clone(ctx); 
         newCtx.body = userClone;
+        // skip events when calling uc.save, so that validate and put is not called from this
+        // internal call
+        newCtx.internalSkipEvents = true;
```
ericfong: add _ as prefix?
_internalSkipEvents

lib/resources/collection/index.js
```diff
@@ -700,7 +700,9 @@ Collection.prototype.execCommands = function (type, obj, commands) {
   return this;
 };
 
-Collection.prototype.shouldRunEvent = function(ev, ctx) {
+Collection.prototype.shouldRunEvent = function (ev, ctx) {
+  if (ctx && ctx.internalSkipEvents) return false;
```
ericfong: Seems hard to maintain

andreialecu:
@ericfong: updated

ericfong:
landed as 81574fcbd523c0c708f54226ad6e47ad64804386
Thanks

By the way, please use a clean branch to do PR. Will be better for rebasing

PR #486, opened 2015-01-14
Fix(session): race condition crash

andreialecu:
Should not crash process when inserting the same session to the database.


andreialecu:
Here's an instance of this bug appearing in the wild: https://travis-ci.org/deployd/deployd/builds/46467404 look for E11000

NicolasRitouet:
Landed as 00b955542950988ef5e7b7c522997d16cfa0f5a5

Thanks @ericfong  :+1:

PR #489, opened 2015-01-16
refactor(session): Various improvements

### LINKED ISSUE #314: Huge session collection
I am using Deployd for a year+, and the problem is that my sessions collection is quite big - is it safe to remove all records from this collections on DB level?

db.sessions.stats()
{
	"ns" : "database.sessions",
	"count" : 14469487,
	"size" : 2083649920,
	"avgObjSize" : 144.00302650674485,
	"storageSize" : 3918532608,
	"numExtents" : 22,
	"nindexes" : 1,
	"lastExtentSize" : 1021419520,
	"paddingFactor" : 1,
	"systemFlags" : 1,
	"userFlags" : 0,
	"totalIndexSize" : 3894482256,
	"indexSizes" : {
		"_id_" : 3894482256
	},
	"ok" : 1
}

If it is safe, maybe there should be mechanism for auto cleaning old records ?

andreialecu:
No longer create sessions on every request if sid cookie is not specified. Fixes #314 
Add server session expiration, defaults to 30 days. Can be overriden like:
```
deployd({
  sessions: { maxAge: 31556952000 /* 1 year */ },
  .. other options ..
```
Automatically invalidate sessions on username/password change.

Note: This PR will clean up the sessions table once implemented in order to prepare it for the new format. This is not a big deal, it will just require your users to log in again.


ericfong:
Hi @andreialecu, I would like to rebase this PR. But found many conflict while rebasing, because some codes in master are changed and conflicted.

I may need your help to update the code to match with the bluebird and other changes

andreialecu:
@ericfong I just updated this to work on current master.

ericfong:
Thanks for your contribution!
Landed as 26b9448ee4c89d4ce5d5b4f1b5eb30b99a125873

isAnonymous is interesting and useful. need documentation.

All testes look good to me.
A bit worry about:
```
process.nextTick(function () {
    store.cleanupInactiveSessions();
});
```
Seems node.js 0.12 and iojs 1.3.0 dislike process.nextTick.
I may be wrong.

PR #493, opened 2015-01-21
fix(script): pairing of addCallback and finishCallback to keep callbackC...

ericfong:
pairing of addCallback and finishCallback to keep callbackCount in right number

### DIFF
--- a/lib/script.js
+++ b/lib/script.js
@@ -182,9 +182,10 @@ function wrapAsyncFunctions(asyncFunctions, sandbox, events, done, sandboxRoot) 
             }
           };
         } else {
-          events.emit('addCallback');
           args.push(function() {
             if (sandboxRoot._error) return;
+            // pairing of addCallback and finishCallback to keep callbackCount in right number
+            events.emit('addCallback');
             events.emit('finishCallback');
           });
         }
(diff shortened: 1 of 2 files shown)


lib/script.js
```diff
@@ -184,6 +184,8 @@ function wrapAsyncFunctions(asyncFunctions, sandbox, events, done, sandboxRoot)
         } else {
           args.push(function() {
             if (sandboxRoot._error) return;
+            // pairing of addCallback and finishCallback to keep callbackCount in right number
+            events.emit('addCallback');
```
andreialecu: Adding this here for reference.

Add then immediately finish doesn't really seem like a good idea. Might as well not even finish it in the first place on the second line. This wouldn't have any effect this way.

Also, this change would break PR #484.

andreialecu:
The test isn't entirely correct by the way. The functionality of `addCallback` and `finishCallback` is supposed to suspend returning from the async function until it is done. You're incrementing `count` synchronously. In order for the test to be relevant you will need to increment `count` from inside the `setTimeout` block and check whether in the end the result is 3.

ericfong:
suspend returning from the Script until ALL async functions within it are done

The problem is ALL. Actually, we don't really need to wait for ALL in this case. 

Anyway, I will close this and hope the promise one go through and solve the same problem

PR #494, opened 2015-01-23
feat(core): wait for promises in events

simpleTechs:
With these changes you can safely use promises in events, because they are now wrapped just like async functions before. This allows for way cleaner code in event handlers.

Also, this change fixes a minor issue where the function `wrapAsyncFunctions` would silently discard every property of the wrapped function. This leads to problems when you put constructors inside a domain that is wrapped (i.e. only the constructor function was retained, static properties and methods were discarded).

### DIFF
--- a/lib/internal-client.js
+++ b/lib/internal-client.js
@@ -1,5 +1,5 @@
-var debug = require('debug')('internal-client'),
-    bluebird = require('bluebird');
+var debug = require('debug')('internal-client')
+  , bluebird = require('bluebird');
 
 function normalizeArray(parts, allowAboveRoot) {
   // if the path tries to go above the root, `up` ends up > 0
@@ -84,9 +84,8 @@ exports.build = function(server, session, stack) {
       var callback = function(data, error) {
         if (typeof fn === 'function') {
           // call our callback
-          deferred.promise.then(function () {
-            fn(data, error);
-          }, function () {
+          deferred.promise
+          .done(function () {
             fn(data, error);
           });
         }
--- a/lib/internal-client.js
+++ b/lib/internal-client.js
@@ -1,4 +1,5 @@
-var debug = require('debug')('internal-client');
+var debug = require('debug')('internal-client'),
+    ayepromise = require('../clib/ayepromise');
 
 function normalizeArray(parts, allowAboveRoot) {
   // if the path tries to go above the root, `up` ends up > 0
@@ -65,7 +66,8 @@ exports.build = function(server, session, stack) {
         , res
         , urlKey
         , recursions
-        , recursionLimit;
+        , recursionLimit
+        , deferred = ayepromise.defer();
 
       req = {
           url: joinPath('/', options.path)
@@ -77,6 +79,23 @@ exports.build = function(server, session, stack) {
         , internal: true
         , headers: {}
         , on: function() {}
+      };
+
+      var callback = function(data, error) {
+        if (typeof fn === 'function') {
+          // call our callback
+          deferred.promise.then(function () {
+            fn(data, error);
+          }, function () {
+            fn(data, error);
+          });
+        }
+        // resolve or reject promise
+        if (error) {
+          deferred.reject(error);
+        } else {
+          deferred.resolve(data);
+        }
       };
 
       urlKey = req.method + ' ' + req.url;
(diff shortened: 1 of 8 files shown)


simpleTechs:
While this may seem like a duplicate of #484 it is, in my opinion, way cleaner (in both detecting a promise and in the way it works, i.e. it's not limited to the internal client).
Actually I am about to release an extension that would duplicate #484, but use the changes proposed in this very PR. 

The idea was to add promise support at the lowest level possible and then allow the user to decide if he wants to have the internal client support promises or not.

andreialecu:
Hey @simpleTechs: The changes in #484 are not limited to the internal client either. If a promise is detected by `wrapAsyncFunctions` it is handled regardless of where it is implemented:
https://github.com/deployd/deployd/pull/484/files#diff-1635b4f24e4bb7c1006e010d10362130R194

Resolving the promise is handled by calling the function injected as the last parameter, from within the wrapped async function. It seems cleaner to me in #484 as less changes are required.

Note that the test in #484 will also not work with this PR, because `this` inside the promiseable won't be  set properly, something https://github.com/petkaantonov/bluebird is needed for.

andreialecu:
@simpleTechs actually, having looked at this into more detail, it is something I wanted to do in #484 as well but I was concerned about chaining promises and one other potential quirks. It can be implemented as a one line change here:
https://github.com/deployd/deployd/pull/484/files#diff-1635b4f24e4bb7c1006e010d10362130R196

```
result.then(function() { events.emit('finishCallback'); }
```

However, there could be issues when you call the internal client with **both** a callback and a promise, eg:

```
dpd.collection.get(id, function (res, err) { .... }).then(function(res) { ... });
```

Though I'm not sure why you would do that

andreialecu:
Alright, having looked it over more, I like it. :) If the test in #484 would pass too after you make the changes to internal client, that would be awesome.

simpleTechs:
Hey @andreialecu thanks for your feedback and your reconsideration!

I went ahead and ported your changes to `internal-client.js` to `ayepromise` (as it is also used for the js client). I also ported your tests over to the PR.
Note that I change the way you accessed `this` in the `then` arguments, because Promises/A+ specifies it otherwise (and we don't need any other external lib this way).

In my opinion this is even better than providing the promise-support for the internal-client as a dedicated module. Plus now we have the very best of both PRs :)

What do you think?

andreialecu:
Yes, I know what the spec says but I think the events are dirty without that feature. What is in the spec isn't supposed to be set in stone anyway. :)

Also, Q is already referenced in package.info and used in several places internally. I'm not sure why you're using ayepromise from clib. I think it would've been better to use Q.

My PR changed from Q to bluebird, which is actually better than Q for a lot of things and brings the feature of `this` support in thenables.

andreialecu:
By the way, I already discussed with @NicolasRitouet and @ericfong and it's ok to switch to bluebird. Was about to merge my PR anyway, but I think we can merge this one instead.

ericfong:
I think we finally (in 0.9 or 0.10) will not wrap the func/callback like now, (which is very magic but create a log of problem). 

Let's fix the problem or keep it dirty. And decide what is more production ready and yet easy to use way

ericfong:
I think we should keep deps as less as possible

BTW, I think we finally (in 0.9 or 0.10) will NOT wrap the func/callback like now, (which is very magic but create a log of problems: vm is slow. It forces event to wait for ALL async calls). Let's fix the problem  quickly and move on

andreialecu:
I think it's actually pretty smart that it waits for async calls this way. We could expose addCallback and removeCallback on the context and manually decide which callbacks are important to wait for, but I don't necessarily think that should be the default.

Regarding "vm is slow", care to elaborate? This shouldn't have any measurable impact actually.

NicolasRitouet:
As Eric said, we have to keep the dependencies as lean as possible.
I have to say, I have a preference for ```bluebird```.
Is it doable in this PR to migrate to this one?
In the frontend, we'll have to keep ```ayepromise``` until somebody make the migration in another PR.

andreialecu:
Regarding the frontend: I think it's ok to keep ayepromise because it's very small and Q and bluebird can take and consume `ayepromise` promises, so if you want to convert them in the front end to full featured that's possible.

ericfong:
For vm, 

http://www.davidmclifton.com/2011/08/18/node-js-virtual-machine-vm-usage/

Not sure still valid. And we are using ```script.runInNewContext```

May be should do a benchmark on this.

andreialecu:
@ericfong According to those benchmarks, even in the worst case it's still **under a quarter of a milisecond per call**. It's doing 5000 iterations in 1110 ms which is not bad at all for our use scenario.

simpleTechs:
As requested and discussed on gitter, I updated the PR to use bluebird instead of ayepromise internally.

ericfong:
Thanks for your contribution!
Landed as 964ec452619e7da7a7f5ce0f87fb6d334dbc21cd

Cannot find any problem so far.
Hope users can help to find issue out.

PR #495, opened 2015-01-26
fix(db): Should not crash process when an invalid $fields query is passed

andreialecu:


NicolasRitouet:
Landed as ca68e1531e5f692e29b9f67f35540faeb7b60a45

Thanks @andreialecu , another splendid PR :+1:

Issue #353, opened 2014-05-02
server very slow


robotnic:
My deployd server is really slow and needs lots of memory.

Here the timing with node and mongodb on localhost with 50 items in collection users.

> time wget http://localhost:3000/users
> real    0m0.474s

If I add an event, everything gets even slower (get.js contains a javascript comment and nothing else)

> vi resources/users/get.js
> time wget http://localhost:3000/users
> real    0m2.343s

---

Also Memory consumption is insane. It adds 10 MByte per request and never frees that.

ps -eo pid,rss | grep 10078
10078 182596
wget http://localhost:3000/users

ps -eo pid,rss | grep 10078
10078 193184

Memory jumps up from 182 MByte to 193 MByte with a single request.
On heroku with mongohq I have the same problem.

What am I doing wrong? Any idea?

Bernhard

PR #496, opened 2015-01-27
fix(script): memory leak and performance improvement

andreialecu:
There is a bug in node with vm.runInNewContext: joyent/node#6552
This was used for every script call by deployd, and for example: GET scripts are being run for every record returned -> so if you need to return 1000 records, the scripts would run 1000 times resulting in big memory leaks.

With this change, even 15000 executions of a script take now just 300 ms on my machine and use virtually no extra memory, compared to huge memory leaks and over several minutes with the old code.


andreialecu:
This will most likely fix: https://github.com/deployd/deployd/issues/353

lib/script.js
```diff
@@ -9,12 +9,34 @@ var vm = require('vm')
...
   }
 }
 
+Script.prototype.runInSandbox = function (sandbox) {
+  var functionArgs = Object.keys(sandbox);
+  
+  // remove the argument 'this' from our list of passed argument, because it is a reserved word
+  functionArgs.splice(functionArgs.indexOf('this'), 1);
+    
+  var script = '(function (' + functionArgs.join(',') + ') { return (function() {' + this.src + '\n}).call(_this); })';
```
ericfong: ```this.sourceCode``` instead of ```this.src``` ?

lib/script.js
```diff
@@ -9,12 +9,30 @@ var vm = require('vm')
 
 function Script(src, path) {
   try {
```
NicolasRitouet: what's the purpose of the try/catch?

andreialecu:
There is a side effect for this:

Because the script code is now running in the same context, it will have access to node globals like `require` automatically. 

The purpose of the sandbox was **not** to isolate code from a security perspective as far as I can tell, because the code is your own code and is not third party/untrusted code. The sandbox was a way to supply a set of predefined *global* methods (error, cancel, emit, etc) and the ability to update the model from the event script without calling specific APIs. 

I believe `runInNewContext` was probably just the most straight forward implementation of implementing this so that's why it was chosen. However, there is another way of doing it, which is what this PR suggests, while also providing access to node globals. 

I think the fact they were inaccessible before was a lack of an important feature, so this can also be considered a new feature/improvement.

lib/script.js
andreialecu: There isn't one any more actually, I'll fix this. :)

NicolasRitouet:
@ritch @jeffbcross @dallonf 
What do you guys think?
Does it make sense?

andreialecu:
Hopefully I didn't overlook anything, all the tests pass and I tested in my own app for a bit and everything is fine. 

Either way, something needs to be done about it. While the sandbox seems like a good idea, it just isn't stable enough in current `node` versions. Maybe it can be reimplemented in the future, but I see no reason for it (again: unless I'm missing something).

dallonf:
The original reason that we restricted Node globals was because of the script wrapper that counts async callbacks and automatically returns; this works well in a controlled environment where the only async functions are part of dpd.js, but could easily get confused if you were to start importing modules from Node-land. 

(Remember, Deployd was originally designed as a tool for front-end developers rather than Node developers, and favored ease of use [and occasionally magic like callback counting] over conformity to Node conventions. This is no longer the design goal, but many architectural decisions were made based on that.)

So I think that this is a good change that better fits Deployd's current goals. But the callback counting feature will need to be clearly documented along with its caveats, and in the long term, it should be possible to disable it on a per-event basis. (I would hesitate to remove it entirely, since it *is* useful most of the time)

ipepe:
Why if it is so big improvement, this pull request isn't pulled?

NicolasRitouet:
We will merge this PR, but we have other PR which are also important and which contains less potential side-effect. We have to prioritize.
As soon as somebody from the team has time, we'll test this PR and eventually merge it.

ericfong:
Testing this PR. 

@andreialecu would you mind update this PR?

Conflict with the Promise commit

andreialecu:
@ericfong fixed.

ericfong:
Thanks for @andreialecu contribution!
Landed as 1f5270e678f622b2bda2366d96e8f1a05acc24c8

As @andreialecu mentioned, 
Because the script code is now running in the same context, it will have access to node globals like require automatically.

Hence, we better bump the version and doc this change. It is not breaking change but may have many side effect. 

And
```
require('../relative_path_may_cause_problem')
```

hongkongkiwi:
I have downloaded the latest version but I still don't have access to require in the current context when running events, any ideas why that is? According to this closure it should just work as a normal var blah = require("blah");

n321203:
I believe require() is lost again – with a clean installation of v0.8.5, the call require() in events results in "require() is not defined" error.

PR #497, opened 2015-01-27
fix(collection): $push should work with arrays when inserting new records

andreialecu:
This is mostly a convenience and for consistency with PUT. It allows to set arrays using the same syntax available in PUT for arrays ($push, $pushAll).

### DIFF
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,0 +1,8 @@
+node_modules/**
+test-app/**
+clib/*.min.js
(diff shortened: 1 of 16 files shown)


andreialecu:
This is very minor and can be committed to 0.8

ericfong:
Landed as 077c2b97cffee6c14f4f10391c3f78b7055aae8d
Thanks

Issue #149, opened 2013-01-29
Angular Routes


scharissis:
Hi,

I can't seem to use angular routing (for non '/' routes). I would like to have an index.html with an ng-view which loads partials/templates based on the url path.

The Deployd collection API defines routes (eg. GET /todos) which I can't use, but even if I try a non-collection route, I'm always directed to a '404 - Resource Not Found' page instead of my template being loaded.

Ideally I'd like "/todos" to show my index+template and move the delpoyd api to /api/todos (https://github.com/deployd/deployd/issues/83), but I'd settle for something like "/listTodos" for now.

I've found that Angular routes kind-of-work: navigating to a url will cause a Deployd REST API and you'll see it's JSON output. But, if you change the path (with locationProvider) in your Controller(s) then it triggers the Angular template to get injected into the ng-view. Any page reloads break it. But, of course, this is less than ideal...

Is it possible to get this actually working? Am I missing something or does the Deployd routing need to be more extensible?

Cheers,
Stefano

PR #490, opened 2015-01-18
feature(dpd.js): dpd script can change its url base

ericfong:
Thanks for

PR #437 by @iranreyes
gist https://gist.github.com/andreialecu/45eba279e4ccc009a5af by @andreialecu 

I have combined them and added a little test to cover it.

Please take a look

PR #498, opened 2015-01-27
fix(dpd.js): parse script src for root url, apposed to window.location

### LINKED ISSUE #83: Defining an API Root
It would be pretty useful if you could define a different point for which API requests would be responded to. For example:

before:
GET /contacts
GET /users

after having defined a different API root: (/api/)
GET /api/contacts
GET /api/users

sscovil:
enable users to host deployd on a subdomain (e.g. api.mydomain.com) and still use dpd.js

might resolve #83 & #149, see https://groups.google.com/forum/#!topic/deployd-users/CT84hulBO40

### DIFF
--- a/clib/dpd.js
+++ b/clib/dpd.js
@@ -2,10 +2,7 @@
 
   if (!window._dpd) window._dpd = {};
 
-  var root = window.location.protocol + '//' + window.location.hostname;
-  if (window.location.port !== '') {
-    root += ':' + window.location.port;
-  }
+  var root = document.currentScript.getAttribute('src').replace(/\/dpd\.js$/, '');
 
   var consoleLog = (typeof console !== 'undefined') && console.log;
 
--- a/clib/dpd.js
+++ b/clib/dpd.js
@@ -21,6 +21,7 @@
       for (var i = 0; i < scriptTags.length; i++) {
         if (scriptTags[i].hasAttribute('src') && scriptTags[i].getAttribute('src').match(/.*dpd\.js$/)) {
           element = scriptTags[i];
+          break;
         }
       }
     }
--- a/clib/dpd.js
+++ b/clib/dpd.js
@@ -2,18 +2,30 @@
 
   if (!window._dpd) window._dpd = {};
 
-  var root = window.location.protocol + '//' + window.location.hostname;
-  if (window.location.port !== '') {
-    root += ':' + window.location.port;
-  }
+  var root = currentScript().getAttribute('src')
+      .replace(/(.*)(\/.*\.js)$/, '$1')
+      .replace(/^[^\/]*.js$/, '');
 
-  var scriptTags = document.getElementsByTagName('script');
-  for (var i = 0; i < scriptTags.length; i++) {
-    var src = scriptTags[i].getAttribute('src');
-    if (src !== null && src.match(/.*\/dpd\.js$/)) {
-      root = src.replace(/\/dpd\.js$/, '');
-      break;
+  function currentScript() {
+    // Only works in modern non-IE browsers
+    var currentScript = document.currentScript;
+
+    if (!currentScript) {
+      // Allows script to have a name other than dpd.js by assigning id="dpd.js" to script tag
+      currentScript = document.getElementById('dpd.js');
     }
+
+    if (!currentScript || !currentScript.hasAttribute('src')) {
+      // As a last resort, loop through all script tags and find the src attribute ending in "dpd.js"
+      var scriptTags = document.getElementsByTagName('script');
+      for (var i = 0; i < scriptTags.length; i++) {
+        if (scriptTags[i].hasAttribute('src') && scriptTags[i].getAttribute('src').match(/.*dpd\.js$/)) {
+          currentScript = scriptTags[i];
+        }
+      }
+    }
+
+    return currentScript;
   }
 
   var consoleLog = (typeof console !== 'undefined') && console.log;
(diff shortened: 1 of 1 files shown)


sscovil:
I know there have been a lot of discussions and a couple of pull requests regarding adding customizable endpoint paths, but this pull is a lightweight fix that simply ensures `dpd.js` can be used in conjunction with a proxy pass in Nginx or Apache.

See the following discussion in the Deployd user group for details: https://groups.google.com/forum/#!topic/deployd-users/CT84hulBO40

sscovil:
After testing on IE, I found that `document.currentScript` is not supported and had to instead use `document.getElementsByTagName('src')`, looping through the results until the `/dpd.js` src is found.

As a result, I decided to leave the original `root` variable declaration in place as a failsafe. If some crazy how my technique fails, root will still have a value. I'm not crazy about that sort of defensive coding, but I didn't want to run the risk leaving root undefined...thoughts?

ericfong:
please take a look 
#490 and
#437

sscovil:
@ericfong I saw those and, while I see value in being able to configure the root directory, I believe the default use of `window.location.protocol`, `window.location.hostname` and `window.location.port` is flawed. Wherever I import `dpd.js` from should, in my opinion, be the default root for API calls.

I think this change could be incorporated into #490, but could be used on its own if that pull won't make it into the next release. What do you think?

ericfong:
I think dpd.js may not called ```dpd.js```

Like me,
My build script download latest dpd.js from server
And compress it with all other codes to build a 'app.bundle.js'.
And I distribute that within my cordova app.

Why?
that is my quick fix to prevent dpd become undefined while mobile is offline.

sscovil:
@ericfong I think yours is a good case for configurable root variables, but I still think the existing code that uses `window.location` is broken and should be fixed sooner than later.

Using the `dpd.js` script src, however it is parsed, is more accurate than the current method which does not allow for proxy-ing an instance of Deployd. If the configurable solution presented in #490 doesn't make it into the next release, this (or some variation of this) should IMHO.

sscovil:
@ericfong My latest commit uses a series of methods to determine the script src and is flexible enough to handle your case of having a different file name.

* First, it attempts to use `document.currentScript` which is the most efficient method, but not supported in IE or older browsers.
* Next, it looks for an element with `id="dpd.js"` that also has a `src` attribute. This is how you can indicate the correct script in your case.
* Finally, it loops through all script tags until it finds the one ending in `dpd.js`. This is least efficient, but will work in any browser.

Once the script element is found, any .js filename is removed from the `src` value (along with the trailing slash) and the remaining string is assigned to `var root`.

clib/dpd.js
```diff
@@ -2,9 +2,31 @@
 
   if (!window._dpd) window._dpd = {};
 
-  var root = window.location.protocol + '//' + window.location.hostname;
-  if (window.location.port !== '') {
-    root += ':' + window.location.port;
+  var root = dpdScriptElement().getAttribute('src')
+      .replace(/(.*)(\/.*\.js)$/, '$1')
+      .replace(/^[^\/]*.js$/, '');
+
```
ericfong: Is it possible that root still not yet set in here? 

No currentScript, no id=dpd.js and script src is changed?

if yes, let's add a checking and fail back to the old code?

ericfong:
BTW, do you think use a function to setup the root Url is better ?

Or use a auto parse way?

We can raise that to 0.8. if more people think that is necessary

clib/dpd.js
sscovil: Good catch. Actually, I'm thinking a polyfill for document.currentScript might be the best way to go here...this looks promising: https://github.com/JamesMGreene/document.currentScript

ericfong: We better don't add dependencies at this stage.

clib/dpd.js
```diff
@@ -2,9 +2,31 @@
...
+  function dpdScriptElement() {
+    // Only works in modern non-IE browsers
+    var element = document.currentScript;
+
+    if (!element) {
+      // Allows script to have a name other than dpd.js by assigning id="dpd.js" to script tag
+      element = document.getElementById('dpd.js');
```
ericfong: This is a bit strange to me. 
As most people won't setup like this id="dpd.js". 

Please remove document.getElementById('dpd.js') to save some processing time.

clib/dpd.js
```diff
@@ -2,9 +2,31 @@
...
+    }
+
+    if (!element || !element.hasAttribute('src')) {
+      // As a last resort, loop through all script tags and find the src attribute ending in "dpd.js"
+      var scriptTags = document.getElementsByTagName('script');
+      for (var i = 0; i < scriptTags.length; i++) {
+        if (scriptTags[i].hasAttribute('src') && scriptTags[i].getAttribute('src').match(/.*dpd\.js$/)) {
```
ericfong: use document.querySelectorAll('script[src$="dpd.js"]') ?

andreialecu:
Another possible idea could be to use an attribute like `data-dpd-host="auto"` or with a URL specified instead of auto, used on the script tag. Auto would mean it is implied from the src.

ericfong:
Actually, I prefer rebase #490  which originally coded by @iranreyes and @andreialecu gist.

That PR should solve this problem by calling js function instead of adding a special to the html

drewdotpro:
Please, for the love of God pull this pull request.
11/10

andreialecu:
I think we'll opt for #490 instead, which incorporates these changes and is more powerful.

drewdotpro:
OK good luck. We moved over to use sailsjs.org as it's (sorry) way better as an API building tool.

andreialecu:
This has been fixed by merging PR #490 which includes the functionality of this PR. Closing this.

PR #499, opened 2015-01-28
fix(attach.js): mkdir resourcesPath if not exists

ericfong:


andreialecu:
Probably the same should be done in server.js

NicolasRitouet:
Landed as 55cf4b366ce951512dac0f06b8fbbcc297c6486a

Thanks.

NicolasRitouet:
> Probably the same should be done in server.js

This PR was specific to the problem I had with Express and Deployd, therefore the fix only in ```attach.js```

PR #500, opened 2015-01-29
fix(tests): call done() properly from async function

andreialecu:
These were causing tests to sometimes fail randomly, there was an inner async function that wasn't properly calling `done()` on the test, so other tests were started before these finished, resulting in resource locks.


NicolasRitouet:
Landed as 061f16de29e1ad81e35b817ce72ff604e61d3a77

Thanks :+1:

Issue #492, opened 2015-01-20
Update Collection does not return Object


hugsbrugs:
Hi,

In one of my modules, I'm updating a collection using http://docs.deployd.com/docs/developing-modules/internal-api/store.html#s-Store.update%28query,%20changes,%20fn%29-1117

```javascript
var WebPages = process.server.createStore('webpages');
WebPages.update({"id":pageId}, webpage, function(err, webpageResult)
{
    if(err)
    {
        // An error occured adding a webpage
        reject(Error("An error occured getting a webpage : " + err));
    }
    else
    {
        console.log("webpageResult : " + webpageResult);
        resolve(webpageResult);
    }
});     
```

Everything runs fine except in the response callback. My webpageResult is always **undefined** but I'm expecting either a boolean or the updated object. Docs are not very clear about it. How do you check query has been successfull because if I give a non existing id for the query, I won't have any error and I'm unable to check it (or I should first do a request to test the id ...)

Thanks for highlighting me !

andreialecu:
I fixed this in PR #501 , it will return an object with the number of records affected, so that  `webpageResult.n` will be 1 in your code above

hugsbrugs:
very nice : thanks @andreialecu

PR #501, opened 2015-01-29
fix(db): update/delete should return count of records affected

andreialecu:
This is specified in the documentation here:
http://docs.deployd.com/docs/developing-modules/internal-api/store.html#s-Store.update%28query,%20changes,%20fn%29-1117
But wasn't working.

It will fix https://github.com/deployd/deployd/issues/492

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -449,9 +449,9 @@ Store.prototype.update = function (query, object, fn) {
   debug('update - command', command);
 
   collection(this, function (err, col) {
-      col.update(query, command, {multi: multi}, function(err) {
+    col.update(query, command, { multi: multi }, function (err, r) {
       store.identify(query);
-      fn(err);
+      fn(err, r ? { n: r.result.n } : null);
     }, multi);
   });
 };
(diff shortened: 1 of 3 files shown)


andreialecu:
This is simple enough to not cause any problems, it also goes together with #502 because of a minor assumption that #502 makes:
https://github.com/deployd/deployd/pull/502/files#diff-e65397a732fc9c867350e8eeed7593fdR386

lib/db.js
```diff
@@ -449,9 +449,9 @@ Store.prototype.update = function (query, object, fn) {
   debug('update - command', command);
 
   collection(this, function (err, col) {
-      col.update(query, command, {multi: multi}, function(err) {
+    col.update(query, command, { multi: multi }, function (err, r) {
       store.identify(query);
-      fn(err);
+      fn(err, r ? { n: r.result.n } : null);
```
ericfong: hesitate to directly use 'n' as field name

andreialecu: What would you suggest? I picked `n` because it made sense for both updates and deletions and wouldn't leak too much info if it got out in the body as a HTTP response (for a DELETE or logout).

ericfong: I am ok with n if for security reason.

Personally, I would like to use 'count' which is easy for documentation

andreialecu:
I just renamed this to `.count`

ericfong:
Thanks for your contribution!
Landed as 6a8caad83e571bf82fdaab3c7bff58cab6d25d42

Issue #488, opened 2015-01-16
please add feature to bulk delete


ramstein74:
hello, this is the way i know to delete all records from collection

dpd.collection.get(function(result,error){
result.forEach(function(entry) {
dpd.collection.del(entry.id,function(){});
});
})

There should be a better way right?

simpleTechs:
While this is most certainly not perfect, what you can do right now is supply a query to `del` - this will not require you to loop over a collection and call it again and again.

Imagine the following scenario: You have an object, say Class, that has many childs called Student. (This relation is setup according to http://docs.deployd.com/docs/collections/relationships-between-collections.html#s-Contains%20Many%20or%20One-to-Many-2191)

So on every Class-instance, you will find an array of "Student-Ids":
```js
{ 
  name: 'Testclass',
  students: [ 'ceb59a120693e8b8', '88acc28d9d7a0973' ],
  id: 'b039e99a43e1f82e'
}
```

Each Student on its own will have some properties, but that's not so important right now:
```js
[
  { name: 'Student A', id: 'ceb59a120693e8b8' },
  { name: 'Student B', id: '88acc28d9d7a0973' }
]
```

What you can do now is simple: construct a query that will select all the students and pass it to the `del` function:
```js
// get them through a query, for example
var studentIds = [ 'ceb59a120693e8b8', '88acc28d9d7a0973' ];

dpd.students.get({ id: { $in: studentIds }}, function(res, err) {
    console.log('get before delete:', res);

    // note: this is only called once, and it will delete more than one object
    dpd.students.del({ id: { $in: studentIds }}, function(res, err) {
        console.log('delete response:', res);
    
        dpd.students.get({ id: { $in: studentIds }}, function(res, err) {
            console.log('get after delete:', res);
        });
    });
});
```

The output will be as following:
```
  get before delete: [ 
    { name: 'Student A', id: 'ceb59a120693e8b8' },
    { name: 'Student B', id: '88acc28d9d7a0973' }
  ]
  delete response: undefined
  get after delete: []
```

This is how you can easily delete in bulk using a query. One thing to keep in mind is that deployd will require an id in the query. This *somehow* makes sense, to limit the outcome of your delete, but there's an easy way to get around this, say you wanted to delete every Student named `Student A`:

```js
dpd.students.get({ name: 'Student A' }, function(res, err) {
    console.log('get before delete:', res);
    
    dpd.students.del({ name: 'Student A' }, function(res, err) {
        console.log('delete response:', err);
    
[...]

PR #502, opened 2015-01-29
fix(collection): call Delete event for every record matched

andreialecu:
You could actually pass a query to DELETE that matches more than one record, but the DELETE event was called just once for the whole operation.
This change correctly allows the event to occur several times, once for each record being matched, and allows canceling removal of certain records via `cancel()` from the event script;

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -383,7 +383,7 @@ Collection.prototype.remove = function (ctx, fn) {
     
     // nothing to delete
     if (result.length === 0) {
-      return fn(null, { n: 0 });
+      return fn(null, { count: 0 });
     }
       
     var remaining = result.length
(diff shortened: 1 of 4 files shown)


andreialecu:
See #488.

andreialecu:
This one is probably important because it's like a SQL injection attack more or less, the original code was supposed to prevent any sort of bulk deletes by forcing the caller to specify an `id`. 

Unfortunately, it overlooked the fact that you could pass something like `{ id: { $ne: null } }` which means that every non null `id` will match. 

This change will make it possible to handle the DELETE event properly without any unexpected side effects. If you want to cancel the deletion, you can still do so, but `this` will be properly populated in the event when the delete is part of a bulk operation.

It shouldn't be a breaking change UNLESS someone relied in the previous (undocumented) behavior.

ericfong:
Thanks for your contribution!
Landed as 462c6766351a1bbac86d730d58f272ca98869527

PR #509, opened 2015-02-09
fix(script): null values in domain were changed to {}

andreialecu:


ericfong:
Thanks for your contribution!
Landed as 26311eea7b0d47bec66f113fca1b1f13bd89ae86

PR #510, opened 2015-02-10
refactor: remove dependency on Q/replace with bluebird

andreialecu:
Bluebird depedency is required by Promise support in event scripts, this commit modifies other places to use bluebird instead of Q


NicolasRitouet:
landed as 1c258b7a25ebf82804a8446b8b2e22f51ae752a1
Thanks @andreialecu  :+1:

PR #512, opened 2015-02-13
Fix: boolean property are correctly set

### LINKED ISSUE #166: Booleans always save as false.
I have a collection that no matter how I submit as true, even the API example booleans save as false.

NicolasRitouet:
fix #166 
Boolean properties used to be set as false.


lib/resources/collection/index.js
```diff
@@ -86,7 +86,7 @@ Collection.prototype.validate = function (body, create) {
         errors[key] = 'is required';
       }
     } else if(type === 'boolean') {
-      body[key] = false;
+      body[key] = val;
```
andreialecu: Leaving this here although we discussed on gitter. This else block is only going to be executed when val is undefined, otherwise the top if block will catch it. 

There was no bug here, the previous version of this file was correct by setting the boolean property to false when an undefined/missing value is passed in.

lib/resources/collection/index.js
```diff
@@ -118,14 +118,16 @@ Collection.prototype.sanitize = function (body) {
...
-    if(expected == actual) {
+    if(expected === actual) {
       sanitized[key] = val;
     } else if (expected === 'array' && Array.isArray(val)) {
       sanitized[key] = val;
-    } else if(expected == 'number' && actual == 'string') {
+    } else if(expected === 'number' && actual === 'string') {
       sanitized[key] = parseFloat(val);
-    } else if(expected == 'string' && actual == 'number') {
+    } else if(expected === 'string' && actual === 'number') {
```
ericfong: seems don't need ===

andreialecu:
I don't think the tests here show the problem because they pass on current deployd master. I have personally haven't encountered this issue so we need a better repro.

ericfong:
I think so

NicolasRitouet:
Indeed, the tests pass without the fix.
I'll come up with the right tests.
Let's close this PR in the meanwhile.

PR #513, opened 2015-02-16
No-cache headers on done()

n321203:
Internet Explorer needs to see no-cache headers, otherwise it will cache the last API response and always return that (ignoring any changes). 

In the current master branch, no-cache headers are not set for PUT:s, which means if you update a collections multiple times (using PUT), IE wont recognize the changes. 

This patch adds no-cache headers to all API responses in the done() function of context.js, so that IE behaves nicely again...

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -92,6 +92,9 @@ Context.prototype.done = function(err, res) {
     }
 
     try {
+      this.res.setHeader("Cache-Control", "no-cache, no-store, must-revalidate");
+      this.res.setHeader("Pragma", "no-cache");
+      this.res.setHeader("Expires", "0");
       if(status != 204 && status != 304) {
         if(body) {
           this.res.setHeader('Content-Length', Buffer.isBuffer(body)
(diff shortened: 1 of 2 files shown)


NicolasRitouet:
Hi @n321203 
Thanks for your PR.
However, it's better for us to have two separate for the two features.
It allows us to:
- discuss separately about the features
- have a cleaner changelog
- have easier eventual rollback
Could you please separate them?

n321203:
Done – this PR now only adds no-cache headers

andreialecu:
+1, seems innocent enough.

lib/context.js
```diff
@@ -92,6 +92,9 @@ Context.prototype.done = function(err, res) {
     }
 
     try {
+      this.res.setHeader("Cache-Control", "no-cache, no-store, must-revalidate");
+      this.res.setHeader("Pragma", "no-cache");
+      this.res.setHeader("Expires", "0");
```
ericfong: Shall we use ```this.res. getHeader``` to check a value have been set or not?

That can increase the flexibility of dpd-module / dpd-plugin

NicolasRitouet: We can add it later in another PR if some people request it.

ericfong:
Thanks for your contribution! @n321203 
Landed as 067e72ac96a15f4b057d058bdc84f5feb5333351

PR #514, opened 2015-02-16
Enable require() in event scripts

n321203:
This patch enables require() in event scripts.

Please note that if you require() and run an asynchronous function in an event script, the event script might finish and return its value without waiting for your async function to complete.

(PS. Never mind the commit messages...)

### DIFF
--- a/lib/script.js
+++ b/lib/script.js
@@ -36,9 +36,6 @@ Script.prototype.run = function (ctx, domain, fn) {
   
   var scriptContext = {
     'this': {},
-    require: function(m){
-    	return require(m);
-    },
     cancel: function(msg, status) {
       var err = {message: msg, statusCode: status};
       throw err;
--- a/lib/script.js
+++ b/lib/script.js
@@ -36,6 +36,9 @@ Script.prototype.run = function (ctx, domain, fn) {
   
   var scriptContext = {
     'this': {},
+    require: function(m){
+      return require(m);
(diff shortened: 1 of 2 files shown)


NicolasRitouet:
Sorry I changed the title, I though you were finished :relaxed:

n321203:
Just went out of battery, now back online :)

lib/script.js
```diff
@@ -36,6 +36,9 @@ Script.prototype.run = function (ctx, domain, fn) {
   
   var scriptContext = {
     'this': {},
+    require: function(m){
+      return require(m);
+    },
```
ericfong: I think we should consider relative modules, like require('../../hello.js')

ericfong: PR #496 may already have this?

n321203:
Yup, seems like #496 already implements this.

lib/script.js
n321203: Yup, seems like #496 already implements this.

andreialecu:
As already noted in a line diff, #496 exposes `require`.

Also, `require` is also available as part of an extension here: https://github.com/mcantelon/dpd-event-extension

n321203:
Closing this, already handled by #496.

projectxmaker:
hi all,

I don't see this patch in master branch ? or I missed sthing ???

PR #521, opened 2015-02-24
feature(server): authenticate with Authorization: Bearer

andreialecu:
Allow authentication using Authorization: Bearer HTTP header instead of cookies. This is especially useful when deployd is used cross domain because cookies are often unreliable.

### DIFF
--- a/lib/attach.js
+++ b/lib/attach.js
@@ -74,7 +74,7 @@ function attach(httpServer, options) {
   }
 
   // persist sessions in a store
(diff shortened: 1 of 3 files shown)


andreialecu:
Cookies still work but if an `Authorization: Bearer <sid>` header is present, the session id will be taken from it,.

Also, if just `Authorization: Bearer` without sid is present in the headers, the `sid` cookie will **not** be returned by the server on a POST to /users/login.

ericfong:
Thanks for your contribution!
Landed as cfe228f926bf3b1e7efd44b0f52edc8ae3323f7b

PR #522, opened 2015-02-25
small fix in dpd keygen

ipepe:
on mac os:
[sudo] npm install deployed -g 
should be run with sudo

### DIFF
--- a/lib/keys.js
+++ b/lib/keys.js
@@ -1,9 +1,9 @@
 var fs = require('fs')
-  , crypto = require('crypto');
+, crypto = require('crypto');
 
 /*!
- * A collection of keys backed by a file.
- */
+* A collection of keys backed by a file.
+*/
 
 function Keys(path) {
   this.path = path || '.dpd/keys.json';
@@ -11,8 +11,8 @@ function Keys(path) {
 module.exports = Keys;
 
 /*!
- * Get a key from the given keys file.
- */
+* Get a key from the given keys file.
+*/
 
 Keys.prototype.get = function(key, fn) {
   this.readFile(function(err, data) {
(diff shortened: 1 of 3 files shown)


ipepe:
Added also dpd keygen fix, when .dpd directory doesn't exist

lib/keys.js
```diff
@@ -41,7 +41,20 @@ Keys.prototype.create = function(fn) {
 
     data[key] = true;
     keys.writeFile(data, function(err) {
-      fn(err, key);
+      if(err){
+        if(err.code == 'ENOENT' && err.path == '.dpd/keys.json'){
+          console.error('Path: '+err.path+' not found, creating new one.');
+          // TODO: Could ask with prompt?
```
ipepe: Should it prompt "Do You want to create a new directory '.dpd'?" if "No such file or directory"?

dallonf:
If you set up Node right (example: nvm), you don't need sudo to use `npm install -g`.

ipepe:
Did You look into keygen fix also?

lib/keys.js
```diff
@@ -41,7 +41,20 @@ Keys.prototype.create = function(fn) {
...
+          fs.mkdirSync('.dpd');
+          keys.writeFile(data,function(error){
```
ericfong: Shall we ensure folder before write?  instead of write > err > create folder > write again ? 

By the way, it seems fix a bug. Just want to know how can we hit this bug?

ipepe: I could implement that. But I dont like whole async/callback implementation of it, so i would implement it in sync methods. I think it would kill it because its just a cmd code. It doesnt need to be async. Right?

ericfong: How can we reproduce this bug (missing of '.dpd/keys.json') ?

Personally, don't find any problem for that

ipepe: Dont use dpd to create a project. Example:
I create a guthub repo, clone it to my projects, npm init, npm i deployd --save, copy start script, dpd keygen-> fail

ipepe:
I refactored code on fixing that directory doesn't exist bug. Do You want me to remove Readme modification or it should stay?

ericfong:
Yes, please remove the Readme.md fix. 

If you can, please also revert the changes of comments of other functions.

Thanks for your contribution

ipepe:
Should be fine now. Also removed changes in Readme file

lib/keys.js
```diff
@@ -21,25 +21,34 @@ Keys.prototype.get = function(key, fn) {
...
-    , keys = this;
+  , keys = this;
 
+  if (this.path.indexOf('.dpd') === 0 && !fs.existsSync('.dpd') ){
+    if( fs.existsSync('node_modules/deployd/.') ){
+      fs.mkdirSync('.dpd');
+    }else{
+      console.error("ERROR! You don't have deployd module in node_modules directory, won't create .dpd directory");
+    }
+  }
```
ipepe: Here is actual changed code now, rest is reformatting code from text editor

andreialecu: Someone may have deployd installed globally with `npm -g` and it won't necessarily be in `node_modules`. I'm not sure why that check is required.

andreialecu:
I think it's better to add a one-liner here instead: 
https://github.com/deployd/deployd/blob/master/bin/dpd#L218

To check if the directory doesn't exist and create it.

(PS: Also, please take care with the reformatting of other code :)

lib/keys.js
ipepe: Well it just won't create a .dpd folder. It doesn't mean that key generation won't take place. It it just a precaution that maybe you typed dpd keygen in folder where You shouldn't have. like in /home/user or /bin? It's precaution for You not paying attention in what folder are you currently. I can change error into some prompt asking: "folder .dpd will be created at path: '/home/user/.dpd' are You sure?"

andreialecu: The prompt sounds good, but see my comment below. I think that's the better place to put this user-land facing code.

ipepe:
I have to disagree with You. In bin/dpd there is no filesystem logic. Also, in lib/keys.js there is variable this.path witch relates to keys path.

andreialecu:
https://github.com/deployd/deployd/blob/master/bin/dpd#L56

ipepe:
I will try to move that code into bin/dpd then...

ipepe:
How about now?

andreialecu:
Thanks for your contribution!
Landed as https://github.com/deployd/deployd/commit/b5a6fd5b5eebdbc755153cc58234459aeabef908

PR #523, opened 2015-02-26
fix(server): allow request.rawBody for middlewares

### LINKED ISSUE #519: Middleware does not work if request stream was already drained
I am using deployd 0.8.0 and recently switched to using deployd as a middleware and created a custom middleware that reads from the request stream on the end event and then calls next. Example code:

```
app.use(require('body-parser')());
app.use(server.handleRequest)
```
probably because the stream already called the 'end' event. There is a related [link](http://lexaux.blogspot.de/2013/10/nodejs-express-middleware-be-aware-of.html).

Maddis1337:
Adds compatibility for middlewares that drained the request stream, e.g. body-parser.
Fixes #519

### DIFF
--- a/lib/util/http.js
+++ b/lib/util/http.js
@@ -67,8 +67,30 @@ exports.setup = function(req, res, next) {
   });
 };
 
+
 /*!
- * Attempts to parse the stream. Currently supports the following formats:
+ * Gets the body the request by either reading a stream or by loading req.rawBody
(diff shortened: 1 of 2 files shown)


ericfong:
Thanks for your contribution!
Landed as 59fbdbab1760bafe043a4f4e4dfef851ac02c7c8

Issue #515, opened 2015-02-16
Wrong error message when deployd crashes


NicolasRitouet:
To reproduce:
```
dpd create test-crash-mongo
cd test-crash-mongo
npm install dpd-fileupload
dpd -d
```
The app crashes because it can't find the resource file from Deployd core and displays an error message about mongodb.
```
Error loading module node_modules/dpd-fileupload
Error: Cannot find module 'deployd/lib/resource'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/Users/nicolas/development/private_workspace/test/node_modules/dpd-fileupload/index.js:6:18)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at /usr/local/lib/node_modules/deployd/lib/type-loader.js:26:32
    at process._tickDomainCallback (node.js:486:13)
Failed to start MongoDB (Make sure 'mongod' are in your $PATH or use dpd --mongod option. Ref: http://docs.deployd.com/docs/basics/cli.html)
```

ipepe:
Did all steps and did not reproduce problem, maybe You have something wrong with "npm install deployd -g"?

ericfong:
It did crash. 
The problem is user forget to "npm install", but the error message think that is MongoDb error

emoleumassi:
i had the same problem. Check this answer. 
http://stackoverflow.com/questions/33120438/deployd-error-cannot-find-module-deployd-lib-resource/33120967#33120967

PR #527, opened 2015-03-05
fix(cli): bug #515 correct wrong error message when deployd crashes

ericfong:
Fix issue #515


NicolasRitouet:
Landed as 59af175c5c40e210828a2aa6ea71cad459d0f8c2

Issue #292, opened 2013-12-13
unique _id as a 10 character alpha-numeric hex for Parse.com migration?


kimballfrank:
Maybe I'm barking up the wrong tree, but I'm very keen on seeing how hard it would be to migrate an existing app from Parse.com to a self-hosted deployd instance. 

One thing out of the shoot is migrating data. Parse unique "objectId" that looks like: "yyuTxK9VDb". Deployd won't accept that string into the "id" field.

ericfong:
Please try to use another field to store the parse.com id
And setup a unique index in mongodb.

PR #528, opened 2015-03-05
fix(collection): POST can specify the item id

ericfong:
For #292

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -128,10 +128,6 @@ Collection.prototype.sanitize = function (body) {
       sanitized[key] = '' + val;
     }
   });
-  
-  if (body.id && typeof body.id == 'string') {
-    sanitized.id = body.id;
-  }
 
   return sanitized;
 };
@@ -499,10 +495,6 @@ Collection.prototype.save = function (ctx, fn) {
 
     store.first(sanitizedQuery, function(err, obj) {
       if(!obj) {
-        if (ctx.req && ctx.req.method == 'POST' && query.id) {
-          // post to create a new item with a specified id and if the obj is not exists before
-          return post();
-        }        
         if (Object.keys(sanitizedQuery) === 1) {
           return done(new Error("No object exists with that id"));
         } else {
@@ -597,23 +589,17 @@ Collection.prototype.save = function (ctx, fn) {
       if(session && session.emitToAll) session.emitToAll(collection.name + ':changed');
     }
   }
-  
-  function validatePost() {
-    if (collection.shouldRunEvent(collection.events.Validate, ctx)) {
-      collection.events.Validate.run(ctx, domain, function (err) {
-        if(err || domain.hasErrors()) return done(err || errors);
-        post();
-      });
-    } else {
-      post();
-    }
-  }
 
   // cannot do (ctx.req.method == 'POST') checking in here as post can also update item
   if (query.id) {
     put();
+  } else if (collection.shouldRunEvent(collection.events.Validate, ctx)) {
+    collection.events.Validate.run(ctx, domain, function (err) {
+      if(err || domain.hasErrors()) return done(err || errors);
+      post();
+    });
   } else {
-    validatePost();
+    post();
   }
 };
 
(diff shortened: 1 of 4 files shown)


lib/resources/collection/index.js
```diff
@@ -589,14 +597,20 @@ Collection.prototype.save = function (ctx, fn) {
       if(session && session.emitToAll) session.emitToAll(collection.name + ':changed');
     }
   }
+  
+  function post() {
```
NicolasRitouet: Could you find a better name? (preparePost? validatePost?)
Between doPost() and post(), I don't know which one is the one to use

andreialecu:
According to http://docs.deployd.com/docs/collections/reference/http.html#s-Updating an Object-1286 objects can be updated via `POST` as well as `PUT` by specifying the `id` in the body. 

Does this PR still allow that behavior?

ericfong:
added a test for post to update case

andreialecu:
I'm still a little concerned about this, since it allows putting anything arbitrary in the id field without any validation.

We need some sort of regex validation at the very least on that field so that power users discovering this implementation detail don't inject arbitrary data as the ID of their records. It can be abused in its current form. 

Perhaps a configuration option such as: 
`deployd({ recordIdFormat: "\\w{5}", ... }) // only 5 character ids allowed`

ericfong:
Only allow 'post' event to set id ?

Disallow for dpd.collection.post({id:'xxx'}) ?

andreialecu:
I think the post event can *already* override the id without any changes in current master. I don't see anything preventing it but I could be wrong. A test for that would be nice.

lib/resources/collection/index.js
```diff
@@ -571,7 +579,7 @@ Collection.prototype.save = function (ctx, fn) {
     if(errs) return done({errors: errs});
 
     // generate id before event listener
-    item.id = store.createUniqueIdentifier();
```
ericfong: Currently, id is always re-generate *after* post event

andreialecu: The comment says *before* right above it though. :)

ericfong:
How about this?

test-app/resources/todos/post.js
```diff
@@ -1,3 +1,7 @@
+if (this.todoId) {
+  this.id = this.todoId;
```
andreialecu: It's not necessary to add `todoId` to the resource. You can just access it by `body.todoId`.

test-app/resources/todos/config.json
```diff
@@ -58,6 +58,14 @@
 			"id": "done",
 			"name": "done",
 			"order": 6
+		},
+		"todoId": {
```
andreialecu: This is not necessary, see my comment in `post.js`

lib/resources/collection/index.js
```diff
@@ -571,7 +571,7 @@ Collection.prototype.save = function (ctx, fn) {
...
+    if (!item.id) item.id = store.createUniqueIdentifier();
```
andreialecu: Not sure why this needs to be changed

ericfong: without this falsy check, item.id will always be a uuid.

It enables the post event to set item.id.

Please follow https://github.com/deployd/deployd/issues/292

ipepe:
Im not sure about this also. Can I somehow turn off this "id post" so some of my users won't exploit it to create some bad ids?

lib/resources/collection/index.js
andreialecu: The `On post` event is called after this `item.id` is populated. What it does is simply prepopulate the id so it is available as `this.id` in the event. 

I think it will be possible to set `this.id` to something else, such as a 10 character alphanumeric id from this `On post` event without any other changes even in current master. Have you verified that you can't?

ericfong:
post event can control the format of id.

If post event file don't has
```
if (this.todoId) {
  this.id = this.todoId;
}
```
this.id should be generated by dpd as usual

ericfong:
Any new comments?

andreialecu:
I don't see what this PR solves. Should we close?

It is already possible to do via the `On POST` event. Simply set this.id = somethingElse (by writing custom logic) there and it should all work.

ericfong:
Actually, id cannot be changed in On Post event. id will be regenerated without this PR.

However, I think dpd should NOT use custom _id now. As that prevent people use monk or other driver to access the database. 

Dpd is not using mongo ObjectId as id. That already create a lot of problems when working with other framework.

PR #530, opened 2015-03-09
fix(internal-resource): handle errors in type events

Maddis1337:
The events used in lib/internal-resource previously ignored any errors that might have occured in the event, because it ignored the error argument in the callback, eventhough the callback might be called with one.
Allows resources to give back errors to the client when an invalid configuration happens.

### DIFF
--- a/lib/resources/internal-resources.js
+++ b/lib/resources/internal-resources.js
@@ -150,7 +150,9 @@ InternalResources.prototype.handle = function(ctx, next) {
         var newId = value.id;
         delete value.id;
         if (typeof value === 'object') value = JSON.stringify(value, null, '\t');
-        notifyType(id, 'Changed', resource, ctx.server, function() {
+        notifyType(id, 'Changed', resource, ctx.server, function(eventError) {
+          if(eventError) return ctx.done(eventError);
+
           fs.writeFile(file, value, function(err) {
             if (err) return ctx.done(err);
 
@@ -257,8 +259,8 @@ InternalResources.prototype.handle = function(ctx, next) {
       var resourcePath = path.join(basepath, 'resources', id);
       wrench.rmdirRecursive(resourcePath, function(err) {
         if (err) return ctx.done(err);
-        notifyType(id, 'Deleted', null, ctx.server, function() {
-          ctx.done();
+        notifyType(id, 'Deleted', null, ctx.server, function(eventError) {
+          ctx.done(eventError);
         });
       });
       break;
--- a/test/resources.unit.js
+++ b/test/resources.unit.js
@@ -193,5 +193,76 @@ describe('InternalResources', function() {
           throw Error("next called");
         });
     });
+
+    it('should call callbacks for config changes with errors', function(done) {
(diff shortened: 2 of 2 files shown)


Maddis1337:
Why I think that this is a bug:
https://github.com/Maddis1337/deployd/blob/internal-resource-event-callback-fix/lib/resources/internal-resources.js#L281

ericfong:
Thanks a lot
But, can you make the CI work too?

Maddis1337:
@ericfong 
The error is not caused by my update. 
It is inside a node module and also you can take a look at my PR and how minor the changes are.

```
/home/travis/build/deployd/deployd/node_modules/doh/lib/handler.js:33
    res.on('close', function () {
        ^
TypeError: Object #<Object> has no method 'on'
    at Domain.<anonymous> (/home/travis/build/deployd/deployd/node_modules/doh/lib/handler.js:33:9)
    at Domain.emit (events.js:95:17)
    at process._fatalException (node.js:255:29)
npm ERR! Test failed.  See above for more details.
npm ERR! not ok code 0

```

andreialecu:
Can you explain exactly what kind of errors were hidden before this fix?

Maddis1337:
Well over functions within deployd use the function with a error object,  see

https://github.com/Maddis1337/deployd/blob/internal-resource-event-callback-fix/lib/resources/internal-resources.js#L281

Other than that I used this in a plug-in I am working on

NicolasRitouet:
I restarted travis ci build, the tests pass now.

andreialecu:
The problem with exposing the actual error text is that something internal/sensitive may be disclosed to a potential attacker. I think we could opt for a more generic text like "Internal server error" instead.

This code is mainly used by the /dashboard however when saving configuration and deleting resources, so I'm not sure what types of errors could occur there.

Care to elaborate how this helps you exactly with your plugin?

Maddis1337:
I develop an extended collection that allows to set indexes like unique to properties in a collection, but without this fix it is impossible to give the user feedback that the index could not get created if he changed the properties.

About the attacker issue: This fix only applies when the user is already authorized to update resources which means he has access to the dashboard.

ericfong:
Do you want to have a PR for "set indexes like unique to properties" to the code of deployd? I think that will be great.

Maddis1337:
Well, this has a pretty severe trade-off, which could be that it is possible that the adding an unique index will fail because the existing data breaks the constraint. This could massively impact existing deployd setups

andreialecu:
@Maddis1337 can you add a unit test for this? It would make it more clear what it adds and it will make sure we won't break it further down the road. It should probably go in `resources.unit.js`.

ipepe:
I support everything that will bring us unique indexing on properties!

ericfong:
You may create another issue for that

Maddis1337:
I finally found time to add unit tests.

andreialecu:
I will have some time this week and will review/merge this then.

andreialecu:
Thanks for your contribution!
Landed as https://github.com/deployd/deployd/commit/7cec886c7adde990b332f1c004d1b761e14d9274

PR #539, opened 2015-03-21
fix(user-collection): do not allow empty passwords

Maddis1337:
it was previously possible to send empty password strings or just the password **null**. 

This should fix that.


andreialecu:
Looks good, although I'm not entirely sure if this was allowed due to a bug or it was a feature. :)

Will review and merge soon unless there are any comments.

ericfong:
Actually, I think some application would like to use empty password.

As some people use dpd as prototyping server.

Maddis1337:
>  Actually, I think some application would like to use empty password.
>  As some people use dpd as prototyping server.

I disagree heavily. Yes for the prototyping server, but it is always a security issue. You cannot let users have accounts without passwords! I would be fine with a toggle that allows this kind of behaviour, but expecting environments to be just prototypes is plain wrong. I know several production instances of deployd.

ericfong:
I am using that for production. And do the password check in my own app.

We can add this feature/bugfix, if noone say no.

andreialecu:
This PR is for 0.7 entirely it looks like, I'm not sure if that was intended. 

I have a new PR at #541 porting this to 0.8.

ipepe:
I think we should have a value 'password length' as fixed atribute of user. Or maybe specify in deployd config minimum password length. Because storing password length is dangerous when database leak occurs.

ericfong:
Any new comments? 
Good to rebase?

NicolasRitouet:
for me, it's good to rebase.

andreialecu:
My only concern is that we already have `required` validation in the core, it might be better to make use of that instead, to keep things more DRY.

andreialecu:
Perhaps it would be a better idea to investigate why this isn't working https://github.com/deployd/deployd/blob/master/lib/resources/user-collection.js#L36

andreialecu:
The alternate fix seems to be changing this 
https://github.com/deployd/deployd/blob/master/lib/resources/user-collection.js#L278
To simply return if no password was passed in, before prepending the salt.

Maddis1337:
any updates?

andreialecu:
@Maddis1337, please see my comment above and see if this works/has the same effect:

```javascript
UserCollection.prototype.setPassword = function (body) {
  if (!body.password) return;
  var salt = uuid.create(UserCollection.SALT_LEN);
  body.password = salt + this.hash(body.password, salt);
};
```

Seems like a better fix for this problem, as the root cause for it is in that method.

Maddis1337:
While this certainly works I think that it is the weaker solution because it will just accept null values without giving back a feedback.

That required does not work is in fact weird. This should return false:
https://github.com/deployd/deployd/blob/master/lib/resources/collection/index.js#L75
according to https://www.npmjs.com/package/validation#validation-exists-value

andreialecu:
@Maddis1337 The point of returning null there is so that the **required** gets to see that it is empty and rejects it properly.

If you look at the setPassword method, even if an empty password is passed to it, `body.password` will still contain *something* after the call -> a salt will be prepended to the empty password. Thus making it contain **something** and pass the *required* check.

Maddis1337:
Ah, now I see what this will achieve. I will take a look into it again.

On Mo., Mai 18, 2015 at 12:31 nachm., Andrei Alecu < notifications@github.com [notifications@github.com] > wrote:
@Maddis1337 [https://github.com/Maddis1337] The point of returning null there is so that the required gets to see that it is empty and rejects it properly.

If you look at the setPassword method, even if an empty password is passed to
it, body.password will still contain something after the call -> a salt will be prepended to the empty password. Thus making
it contain something and pass the required check.

—
Reply to this email directly or view it on GitHub
[https://github.com/deployd/deployd/pull/539#issuecomment-103011334] .[https://github.com/notifications/beacon/ABWATJh1wHPYzHl6zd2MiWf38dpsOUb0ks5oKbcFgaJpZM4Dydbl.gif]

Maddis1337:
So I now spend far too much time for this issue and trying to get the cleaner version to work, because I noticed that there are several issues at play here. Most importantly it has to do with the use of ```sanitize()``` of objects on PUT which does not allow null (because typeof null === 'object'), so empty properties will be deleted. I managed to kind of get it to work by adding a special null case here:
https://github.com/deployd/deployd/blob/master/lib/resources/collection/index.js#L127

But this only works if these two lines are gone:
https://github.com/deployd/deployd/blob/master/lib/resources/user-collection.js#L129-L130

But that breaks two tests. And that is the point where I have to stop for now, because this eats up my whole schedule. Maybe somebody who is more familiar with the process knows how to elegantly solve the underlying issue, that it seems that ```null``` values not possible

Maddis1337:
So of course I had the idea to solve this right after deciding to ditch it...
Anyhow. I actually had an error in my tests because I tried to update a user without login in first. 
This also solves an issue with null values in general.

andreialecu:
Thanks for your contribution @Maddis1337 !
Landed as https://github.com/deployd/deployd/commit/fa07655318f89e1793a0d12772f1e165421cada9

The null value fix may be a breaking change for some users, if they relied on the broken behavior. I have been planning to fix it for a while, so I guess this is a good time.

Hope we won't need to revisit this.

PR #540, opened 2015-03-23
feature(session): bind socket to session

### LINKED ISSUE #533: Emit: unstable for emit to specific user
There seems to be a recurring issue with emit:

I am emitting data to one specific user (for messaging service) using:
```javascript
emit(dpd.users, {id:..}, "action", this)
```
which works, but only now and then! Very arbitrary when it does work and when it doesn't

Also it seems that after some minutes of inactivity the connection is lost, and then the emit stops working. A refresh does give me the new messages so that is not the problem (dpd 0.8.0).

Thanks for the help.

andreialecu:
This allows binding the socket to a session at a later time than the initial handshake when a sid cookie may not be available.

Example usage:
```javascript
dpd.users.login({ username: 'foo', password: 'bar'}).then(function(res) {
  dpd.socket.emit('server:setSession', { sid: res.id });
}
```

It should fix #533

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -44,6 +44,20 @@ function SessionStore(namespace, db, sockets, options) {
         socketIndex[sid] = socket;
         socketQueue.emit(sid, socket);
       }
+      
+      // Alternative way of binding session to socket connection
+      // for when the sid cookie is not yet available.
+      // This expects that the client emits an event with the sid.
+      socket.on('server:setSession', function(data) {
+        if (!data || !data.sid) { return; }
+        
+        var sid = data.sid;
+        var session = sessionIndex[sid]; 
+        if (session) {
+          socketIndex[sid] = socket;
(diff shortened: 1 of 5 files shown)


ericfong:
Shall we patch the User collection login function to emit 'server:setSession' automatically?

andreialecu:
I was thinking about it, we need to patch it in dpd.js (client) somehow. The login function needs special consideration, right now it's just a plain REST call like everything else.

We need to intercept just that one and emit that message from it. I will think of a way to do this cleanly, but it can be a separate PR.

ericfong:
Great!

ericfong:
Thanks for your contribution!
Landed as c4514784558a9ab360537d76dac7704d068bd16a

andreialecu:
Leaving this note here:

@ericfong PR #572 adds calling server:setSession automatically after login.

PR #536, opened 2015-03-16
bugfix: crashes if no password is defined on login

docnoe:
Deployd crashes when someone is sending a login request without specifying a "password" property.
This tiny change fixes this issue by checking if a password is defined at all before checking password validity.

PR #541, opened 2015-03-23
fix(user-collection): crash when password is not specified in login

andreialecu:
Added some further checks for returning an error when either the username or password is not specified on `login`.

### DIFF
--- a/lib/resources/user-collection.js
+++ b/lib/resources/user-collection.js
@@ -162,6 +162,14 @@ UserCollection.prototype.handleLogin = function (ctx) {
     , credentials = ctx.req.body || {};
   
   debug('trying to login as %s', credentials.username);
+  /* jshint eqnull:true */
+  // disable the jshint warning about needing === below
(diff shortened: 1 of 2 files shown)


andreialecu:
Originally reported as #536

Maddis1337:
I think you should keep/add my test cases: https://github.com/deployd/deployd/pull/539/files#diff-d2c37750d7b8fa629df78ceee2b16e6cR368

andreialecu:
@Maddis1337 your PR is separate from this, mine is about `login`, yours is about creating users. Will merge yours separately.

Maddis1337:
@andreialecu I deleted my comment, it was because you commented yesterday (which you later deleted). I will delete this comment later today, too.

NicolasRitouet:
landed as 65f4170b7881ebb01ba50c4ee47fae22f3fe001e

Thanks @andreialecu  :+1:

PR #542, opened 2015-03-24
fix(dpd.js): dpd.socket shouldn't be null

andreialecu:
dpd.socket was null because of: https://github.com/deployd/deployd/commit/302b93d8a5b35b878375eeaca62767ecd386609f

Moved some things around to fix the failing test case.

### DIFF
--- a/clib/dpd.js
+++ b/clib/dpd.js
@@ -4,70 +4,13 @@
 
   var root = null;
 
-  setBaseUrl();
+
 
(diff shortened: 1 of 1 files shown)


NicolasRitouet:
Rebased (5331ac397400e0c6c11a6caee3ea1d1d3cefa58b)
Thanks @andreialecu :+1:

PR #548, opened 2015-04-10
fix(dashboard): use ejs delimiter instead of open and close tag

ericfong:


NicolasRitouet:
Looks good, thx.
@andreialecu do you have time to rebase this one?

andreialecu:
What was the background behind this change? I think I missed it.

ericfong:
last rebased PR upgraded the ejs version and ejs interface is changed.

So this fix to use the new ejs interface.

PR #553, opened 2015-04-21
Delete changed events.

huseyinkozan:
To be able to use 'changed' events at our custom event filters.

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -407,7 +407,6 @@ Collection.prototype.remove = function (ctx, fn) {
 
       if(!remaining) {
         store.remove({ id: { $in: idsToDelete } }, fn);
-        if (session.emitToAll) session.emitToAll(collection.name + ':changed');
       }
     }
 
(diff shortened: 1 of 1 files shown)


andreialecu:
I'm in favor of removing these built in `colname:changed` events because anyone can easily listen in on socket.io and notice which collections are being changed in the database. It can open all sorts of side-channel attacks, for example by someone seeing which collections are changed in response to a certain action.

Removing them may be a breaking change for some apps that are relying on these, but it can be quickly fixed by manually adding the `emit('colname:changed')` in the POST/PUT/DELETE events on a per collection basis.

Paging @NicolasRitouet @ericfong for input.

ericfong:
+1 for remove colname:changed

I don't like "by default" event push to all clients too.

NicolasRitouet:
I'm also ok for changing this.
We're not stable yet, breaking changes are completely ok.

andreialecu:
I have a more complete implementation at #555. I will be closing this, moving discussion there.

Thanks!

PR #554, opened 2015-04-22
fix(collection): should not crash on deleting non existent id

andreialecu:


ericfong:
Thanks for your contribution!
Landed as 1b85c2127c0badda706e34e10522e20dbd7bd879

PR #558, opened 2015-05-06
fix(store): add error handling in callbacks

chcokr:
Add error handling inside certain callbacks in Store's methods. Despite an error being passed in and therefore no second argument such as `col` is being passed in, these callbacks assume the second argument is not `undefined` and make use of it, causing ReferenceErrors. Affected methods are as follows:
- Store.protoype.find
- Store.protoype.first
- Store.protoype.insert
- Store.protoype.update
- Store.protoype.remove
- Store.protoype.rename

No breaking changes, no GitHub issues being closed.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -259,6 +259,10 @@ Store.prototype.insert = function (object, fn) {
   var store = this;
   this.identify(object);
   collection(this, function (err, col) {
+    if (err) {
+      fn(err);
+      return;
+    }
     col.insert(object, function (err, result) {
       if (err) {
         fn(err);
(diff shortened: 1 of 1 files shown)


chcokr:
Feedback appreciated. I don't think this PR is anything difficult.

NicolasRitouet:
Indeed, it's actually so obvious that it looks like it was on purpose.
@andreialecu @ericfong what do you think?
Looks good to me.

chcokr:
Ah I understand. I hope it wasn't on purpose though- DPD crashed for me because those errors weren't being caught :)

ericfong:
I think this PR is good for me. 

May open security issues for dpd modules?
I think even we hide all errors, security issues exist.
Dpd users should understand the code and module that they use.

chcokr:
The code before this PR didn't seem to be much about "hiding" errors. `col` keeps being undefined in case of errors, and the app just crashes. I'm actually surprised DPD doesn't crash for you guys without this PR.

Fundamentally, not handling `err`s in callbacks is like a `throw` not unwinding the call stack. It is a bug, and it is bad practice.

*Not* revealing errors is more of a security issue. I don't believe this PR introduces any new security issue.

Let's get this merged, guys.

NicolasRitouet:
@chcokr thanks for you contribution ! :+1:

PR #557, opened 2015-05-01
fix(session): socket disconnection after login

xroussel:
Issue #533

Issue #141, opened 2012-12-26
emit() with query to multiple connected clients for a single user


justinwalsh:
When using emit and the follow syntax:

emit(dpd.users, {'id': {$in: this.users}}, 'something:happened', this);

The event is only emitted to last connected client by each of the users in the this.users array. For example, if there are two browsers windows open with the same user logged into both, only the browser that is connected last will receive events for that user.

If the query on the emit statement is omitted, then the event is sent to all browsers as expected, but no longer filtered to a group of users.

shaggybb:
There is any plan to solve it?

shaggybb:
This issue has been resolved in 0.8.4 version, hasn't it?

andreialecu:
Yep.

PR #567, opened 2015-05-19
emit() with query to multiple connected clients for a single user (#141)

shaggybb:
This PR allows to store multiple session by user (#141). Also, it keeps in mind pr #540 and #557.

Pending changes:
* getSession method - It needs an extra parameters to get session from array.
* test of getSession - Is necessary to send session identifier.


andreialecu:
Looks like the build is failing here. Also, I merged #557 so you may need to rework this on current master.

shaggybb:
@andreialecu you are right!!

I reverted my fork to create a new pull request.

shaggybb:
@andreialecu new pull request passing tests and merged with #557

PR #569, opened 2015-05-19
fix(session): emit() with query to multiple connected clients for a single user

shaggybb:
tested with mocha && cd test-app && node runtests.js

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -121,9 +121,7 @@ SessionStore.prototype.createSession = function(sid, fn) {
       if (sid) sessionIndex[sid] = sess;
       // index sessions by user
       if (s && s.uid) {
-        if (userSessionIndex[s.uid] === undefined) {
-          userSessionIndex[s.uid] = [];
-        }
+        userSessionIndex[s.uid] = userSessionIndex[s.uid] || {};        
         userSessionIndex[s.uid][sess.data.id] = sess;
       }
       if (!sess.data.anonymous && (!sess.data.lastActive || sess.data.lastActive < Date.now() - 10 * 1000)) {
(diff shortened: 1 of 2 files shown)


lib/session.js
```diff
@@ -293,7 +295,10 @@ Session.prototype.save = function(fn) {
         if(session.rawSocket && session.store.socketIndex)
           session.store.socketIndex[sid] = session.rawSocket ;
         if (res.uid) {
-          userSessionIndex[res.uid] = session;
+          if (userSessionIndex[res.uid] === undefined) {
+            userSessionIndex[res.uid] = [];
```
andreialecu: Shouldn't this be  `userSessionIndex[res.uid] = {};` (aka object, not array?)

Also, the line above should use the more standard way of checking for undefined `typeof .... === "undefined"`

andreialecu:
This needs some unit tests as well.

lib/session.js
```diff
@@ -121,7 +121,10 @@ SessionStore.prototype.createSession = function(sid, fn) {
       if (sid) sessionIndex[sid] = sess;
       // index sessions by user
       if (s && s.uid) {
-        userSessionIndex[s.uid] = sess;
+        if (userSessionIndex[s.uid] === undefined) {
+          userSessionIndex[s.uid] = [];
```
andreialecu: Same here, {} not [].

andreialecu:
By looking at this, I don't think it will work if the user has the same sessionid in a separate tab. It might work across multiple sessions, but not the same one.

I think the main change needs to be done around here somewhere:
https://github.com/deployd/deployd/pull/569/files#diff-0a6235c4445253f49367a39ab5ff3b7cR44

By storing the new connected sockets in an array.

shaggybb:
This code only resolves multiple session for single user. For instance, one user logged in a web browser and a mobile app.

It is true that I haven't solved multiple sockets for the same sessionid. It needs some extra work.

andreialecu:
@shaggybb I have a better fix for this + some unit tests at https://github.com/deployd/deployd/pull/572

Could you please test it in your project and see if it works for you?

It will also work with multiple tabs using the same sessionid hopefully.

shaggybb:
@andreialecu yesterday I was working in something similar to you and I think this is the way to have multiple sockets by sid.
Today I will do some test and I will send you the feedback

andreialecu:
Closing this because #572 was merged, which I based partly on @shaggybb's work in this PR. 

Thanks @shaggybb !

PR #572, opened 2015-05-19
feature(session): emitToUsers() - allow multiple sockets per session AND add options.origins

andreialecu:
Also allows multiple sessions for an user to work properly with emitToUsers().

Improves https://github.com/deployd/deployd/pull/569 and replaces https://github.com/deployd/deployd/pull/557

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -224,11 +224,7 @@ function Session(data, store, sockets, rawSockets) {
     collection.get(query, function(users) {
       var userSession;
       if(users && users.id) {
-        userSession = userSessionIndex[users.id];
-        if(userSession && userSession.socket) {
-          userSession.socket.emit(event, data);
-        }
-        return;
+        users = [users]; // convert single item to array
       }
       users.forEach(function(u) {
         var userSessions = userSessionIndex[u.id];
--- a/lib/session.js
+++ b/lib/session.js
@@ -15,7 +15,7 @@ var sessionIndex = {}
   , userSessionIndex = {};
 
 /**
-* A store for persisting sessions inbetween connection / disconnection. 
+* A store for persisting sessions inbetween connection / disconnection.
 * Automatically creates session IDs on inserted objects.
 */
 
@@ -28,43 +28,46 @@ function SessionStore(namespace, db, sockets, options) {
   // socket queue
   var socketQueue = this.socketQueue = new EventEmitter()
     , socketIndex = this.socketIndex = {};
-  
+
   // NOTE: we will get a warning otherwise when more than 10 users try to login
   socketQueue.setMaxListeners(0);
-    
+
   if(sockets) {
-    sockets.on('connection', function (socket) {
+    sockets.on('connection', function (client) {
       // NOTE: do not use set here ever, the `Cookies` api is meant to get a req, res
       // but we are just using it for a cookie parser
-      var cookies = new Cookies(socket.handshake)
+      var cookies = new Cookies(client.handshake)
         , sid = cookies.get('sid');
 
       if(sid) {
-        // index sockets against their session id
-        socketIndex[sid] = socket;
-        if(sessionIndex[sid])
-          sessionIndex[sid].rawSocket = socket ;
-        socketQueue.emit(sid, socket);
+        var session = sessionIndex[sid];
+        if (session) {
+          // index sockets against their session id
+          socketIndex[sid] = socketIndex[sid] || {};
+          socketIndex[sid][client.id] = client;
+          socketQueue.emit(sid, client);
+        }
       }
-      
+
       // Alternative way of binding session to socket connection
       // for when the sid cookie is not yet available.
       // This expects that the client emits an event with the sid.
-      socket.on('server:setSession', function(data) {
+      client.on('server:setSession', function(data) {
         if (!data || !data.sid) { return; }
-        
+
         var sid = data.sid;
-        var session = sessionIndex[sid]; 
+        var session = sessionIndex[sid];
         if (session) {
-          socketIndex[sid] = socket;
-          socketQueue.emit(sid, socket);
+          socketIndex[sid] = socketIndex[sid] || {};
+          socketIndex[sid][client.id] = client;
+          socketQueue.emit(sid, client);
         }
       });
     });
-  } 
+  }
 
   Store.apply(this, arguments);
-  
+
   if (db) {
     // Cleanup inactive sessions from the db
     var store = this;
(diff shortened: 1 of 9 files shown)


lib/session.js
```diff
@@ -28,43 +28,46 @@ function SessionStore(namespace, db, sockets, options) {
...
-        // index sockets against their session id
-        socketIndex[sid] = socket;
-        if(sessionIndex[sid])
-          sessionIndex[sid].rawSocket = socket ;
-        socketQueue.emit(sid, socket);
+        var session = sessionIndex[sid];
+        if (session) {
+          // index sockets against their session id
+          socketIndex[sid] = socketIndex[sid] || {};
+          socketIndex[sid][client.id] = client;
```
shaggybb: To understand that ¿Is client.id a property of socket?

shaggybb:
First feedback:
I've copied this two files in my project and after passing the tests, I started my project and any change through dpd.collection.on does not trigger an event.

It is very rare, I will check code with node-inspector to understand where is the problem

lib/session.js
andreialecu: Yes, it's the socket.io socket id.

andreialecu:
@shaggybb I pushed a new commit with some further changes.

PS: You can use `npm install andreialecu/deployd#refactor-session2` to use my branch/PR and test instead of manually copying files

shaggybb:
@andreialecu In general everything works great excepting in case when the app emits filtering by only one item. 

Tests Round 1:
Test 1 (Chrome: [Tab + Duplicate tab]):
- Open chrome and loggin with user A.
- Duplicate tab (it duplicates session of tab)
Check 
```javascript
var users = [me.id]; // user A
emit(dpd.users, {id:{$in:users}}, "messages:created", this); // FAIL
emit(dpd.users, {id:me.id}, "messages:created", this);  //SUCCESS
```
Test 2 (Chrome: [Tab + Duplicate tab + new tab]):
- Open chrome and loggin with user A.
- Duplicate tab (it duplicates session of tab)
- New tab and set location
Check 
```javascript
var users = [me.id]; // user A
emit(dpd.users, {id:{$in:users}}, "messages:created", this); // FAIL
emit(dpd.users, {id:me.id}, "messages:created", this);  //SUCCESS
```
Test 3 (Firefox + Chrome:[Tab + Duplicate tab + new tab]):
- Open chrome and loggin with user A.
- Duplicate tab (it duplicates session of tab)
- New tab and set location
- Open firefox and loggin with user A.
Check 
```javascript
var users = [me.id]; // user A
emit(dpd.users, {id:{$in:users}}, "messages:created", this); // FAIL
emit(dpd.users, {id:me.id}, "messages:created", this);  //SUCCESS
```

Test 4 (Firefox + Chrome: [Tab + Duplicate tab + new tab] + Chrome in Private mode):
- Open chrome and loggin with user A.
- Duplicate tab (it duplicates session of tab)
- New tab and set location
- Open firefox and loggin with user A.
- Open chrome in private mode and loggin with user B
Check 
```javascript
var users = [me.id];  // user A
emit(dpd.users, {id:{$in:users}}, "messages:created", this); // FAIL
emit(dpd.users, {id:me.id}, "messages:created", this);  //SUCCESS
```

I've checked that when emitToUser get only only user it does not emit per socket.
Solution:
```javascript
  this.emitToUsers = function(collection, query, event, data) {
    collection.get(query, function(users) {      
      var userSession;
      if(users && users.id) {
        users = [users]; // Converts from single object to array of object. This avoids code repetition and allow to iterate sockets with next lines of code.
        // userSession = userSessionIndex[users.id];
        // if(userSession && userSession.socket) {
        //   userSession.socket.emit(event, data);
        // }
        // return;
      }
      users.forEach(function(u) {
        var userSessions = userSessionIndex[u.id];        
        // emit to sessions online
        for (var key in userSessions) {
          userSession = userSessions[key];
          if (userSession && userSession.socket) {
            userSession.socket.emit(event, data);
          }
        }
      });
    });
  };
```

After this change everything works perfectly!
Also I've tested general emit and works perfectly.

Really really good job!

andreialecu:
@shaggybb You are right, I just pushed the fix. Thanks!

shaggybb:
@andreialecu, could you tell me when this PR will be present in the master or in a near version?

Thanks for your time!

andreialecu:
**Important Note: This PR also strictly rejects unknown origins from receiving responses from deployd. Until now all origins were accepted.**

Use something like:

```javascript
var server = deployd({
  port: process.env.PORT || 1337,
  env: 'production',
  origins: [ "http://localhost:3000", "http://myproductionsite.com" ],
  db: { connectionString: process.env.MONGO_URI }
});
```

In your deployd configuration from now on to whitelist your allowed origins.

**If you do not need to use CORS (you use the same origin) you are unaffected by this.**

ericfong:
how about from cordova app or curl?

andreialecu:
@ericfong What do you mean? If you're referring to the origin fix: 

This is simply a CORS option which web browsers respect to prevent cross site request forgery. Curl or other download methods should be unaffected by this.

NicolasRitouet:
Looks good, I'll merge this, anybody against it?

lib/session.js
```diff
@@ -150,8 +158,8 @@ SessionStore.prototype.createSession = function(sid, fn) {
 /**
 * Get the already created session
 */
-SessionStore.prototype.getSession = function (uid) {
-  return userSessionIndex[uid];
+SessionStore.prototype.getSession = function (uid, sid) {
```
ericfong: May be a breaking change for some code.

lib/util/http.js
```diff
@@ -8,12 +8,12 @@ var Cookies = require('cookies')
  * A utility for setting up a request and response.
  */
 
-exports.setup = function(req, res, next) {
+exports.setup = function(options, req, res, next) {
```
ericfong: May become a breaking change

lib/session.js
andreialecu: Previous implementation was broken, a single uid can have multiple sessions. New signature is more correct.

lib/util/http.js
andreialecu: http.js is not documented as a public API, no breaking change here I think.

lib/session.js
ericfong: Yes. I also support this change. Just want to remark it and let people know.

ericfong:
Thanks for your contribution!
Landed as f67ee1ecc59fd8825059793462fab40fc89d2faf

Issue #565, opened 2015-05-15
User `onGet` event can break session handling


0ff:
This code here https://github.com/deployd/deployd/commit/26b9448ee4c89d4ce5d5b4f1b5eb30b99a125873#diff-a030694f2d36b54385439136636f4a74R101 results in a severely bad situation: The UserCollection runs `getUserAndPasswordHash` on an object that is returned by the `onGet`-Event on that UserCollection.

Think of the following event code:
```javascript
    hide(username);
```
-> getUserAndPasswordHash in this code would return a wrong result, the user would be logged out.

Internal code must not rely on the user's events output.

Maddis1337:
+1

andreialecu:
PR with fix at #574

PR #574, opened 2015-05-21
fix(UserCollection): `onGet` event can break session handling

andreialecu:
Fixes https://github.com/deployd/deployd/issues/565

### DIFF
--- a/lib/resources/user-collection.js
+++ b/lib/resources/user-collection.js
@@ -98,14 +98,25 @@ UserCollection.prototype.handle = function (ctx) {
           return noSuchUser();
         }
 
-        ctx.query = {id: ctx.session.data.uid};
+        ctx.query = ctx.query || {};
+        ctx.query.id = ctx.session.data.uid;
 
-        return this.find(ctx, function(err, user) {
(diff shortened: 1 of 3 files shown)


lib/resources/user-collection.js
```diff
@@ -98,14 +97,25 @@ UserCollection.prototype.handle = function (ctx) {
...
 
-        ctx.query = {id: ctx.session.data.uid};
+        ctx.query = ctx.query || {};
+        ctx.query.id = ctx.session.data.uid;
+
+        // raw store query, need unmodified username and password
+        return uc.store.find({id: ctx.session.data.uid, $fields: {username: 1, password: 1}}, function(err, user) {
+          if (err) {
+            console.error("Internal Error: " + JSON.stringify(err));
+            return ctx.done({statusCode: 500, message: "Error retrieving user for verification"});
```
andreialecu: I'm hiding the actual error here from the response for security purposes as it may potentially contain sensitive data.

NicolasRitouet:
landed as 74f2363141defd66b969a8ad84e2a0932f5c57ff

PR #577, opened 2015-05-25
refactor: compatibility with node.js 0.12

andreialecu:
Various changes to support node 0.12. Updated some dependencies.

Socket.io has some issues with an outdated version of the `ws` package not building correctly on node 0.12 which is only in the master on git, so using that for now. Once they release to npm we can remove the pinned dependencies.


andreialecu:
Ref for socket.io issue: https://github.com/Automattic/socket.io/issues/2056

andreialecu:
Also note: the core of deployd itself was compatible, the main problems were in the tests/mock of timers and runtests.js for the integration tests.

There was also some sort of issue with a test failing because of the `request` package (should've been fixed here: https://github.com/request/request/issues/1421 but apparently it isn't), which I substituted with a plain `http.get()` call: https://github.com/deployd/deployd/pull/577/files#diff-3c1011bb85afc606effe3bfd1af8c66fR41

NicolasRitouet:
Landed as 45d200e216c36092d115fa6647bd006115b4a982
I created an issue to remember to modify the version once the fix is released:
https://github.com/deployd/deployd/issues/578

Thanks @andreialecu

PR #580, opened 2015-05-26
fix(session): session rehydration via socket.io

andreialecu:
When deployd crashes or is stopped and restarted, all the sessions in the memory store are cleared, so when you reconnect via socket.io, it can't associate the connection with a valid session. Previously, all new emitToUsers messages targeting that sesssion were lost until you logged in again so your session got recreated.

This PR fixes this issue so that the session is rehydrated appropriately if a socket.io connection requests it.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -23,6 +23,8 @@ var sessionIndex = {}
 */
 
 function SessionStore(namespace, db, sockets, options) {
+  var self = this;
+
   this.sockets = sockets;
   this.options = options || {};
   // sessions inactive for longer than this will be cleaned up:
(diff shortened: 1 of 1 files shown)


ericfong:
Do we want to have some test over it?

andreialecu:
It's hard to test because deployd needs to be restarted for this fix to apply. It can be mocked probably but it's pretty straight forward. Perhaps later on we can add a test.

ericfong:
Thanks for your contribution!
Landed as 29d22cf2ec3818ce03d19ab9f694a193224b3686

PR #585, opened 2015-06-01
fix(UserCollection): don't allow non string value as password

neojp:
The **user collection** uses `crypto` for password hashing and `crypto.createHash().update()` only allows to be passed strings. While the `setPassword()` method allowed truthy values to be passed over to crypto resulting on a thrown error which stops the server process.

This PR adds extra validation for string values and informs the API user about it.

```
POST /users TypeError: Not a string or buffer
    at TypeError (native)
    at Hmac.Hash.update (crypto.js:97:16)
    at UserCollection.hash (/usr/lib/node_modules/deployd/lib/resources/user-collection.js:343:44)
    at UserCollection.setPassword (/usr/lib/node_modules/deployd/lib/resources/user-collection.js:333:31)
    at UserCollection.handle (/usr/lib/node_modules/deployd/lib/resources/user-collection.js:139:12)
    at /usr/lib/node_modules/deployd/lib/router.js:82:22
    at doNTCallback0 (node.js:408:9)
    at process._tickDomainCallback (node.js:378:13)
```

### DIFF
--- a/lib/resources/user-collection.js
+++ b/lib/resources/user-collection.js
@@ -131,16 +131,23 @@ UserCollection.prototype.handle = function (ctx) {
       }
       /* falls through */
     case 'PUT':
+      function done(err, res) {
+        if (res) delete res.password;
+        ctx.done(err, res);
+      }
+
       this.setPassword(ctx.body);
+
+      // only allow strings on passwords
+      // undefined passwords will be handled by the required validation
+      if (typeof ctx.body.password !== 'string' && typeof ctx.body.password !== 'undefined') {
+        return done({errors: {password: 'is invalid; must be a string'}});
+      }
+
       var isSelf = ctx.session.user && ctx.session.user.id === ctx.query.id || (ctx.body && ctx.body.id);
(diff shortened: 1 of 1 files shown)


Maddis1337:
I think this is a duplicate of #539 and already fixed

neojp:
@Maddis1337 That is definitely the change that introduced this edge scenario. The problem is that by passing a truthy value crypto crashes the process.

```
dpd.users.post({
  username: 'qwerty',
  password: true
});

dpd.users.post({
  username: 'qwerty',
  password: {}
});

dpd.users.post({
  username: 'qwerty',
  password: 123456
});
```

andreialecu:
@neojp would you mind adding a test for this please?

andreialecu:
I think `true` will be handled by this line `if(!body || !body.password || body.password.length < 1) {`, but the empty object `{}` might still pass that check.

andreialecu:
Fixed by https://github.com/deployd/deployd/commit/e81c6c7bc3b28c1c2fddee8c85014724356ca8bb

PR #591, opened 2015-06-28
feature(script): $addCallback $finishCallback

andreialecu:
Add manual callback handling, for when you need to `require` other node functions that do not play well with deployd's automatic async callback handling.

Call `$addCallback()` before calling an async function to increment the counter, and `$finishCallback()` when the callback is finished to decrement the counter. When the counter is 0, deployd will return the response.

### DIFF
--- a/lib/script.js
+++ b/lib/script.js
@@ -44,6 +44,7 @@ Script.prototype.run = function (ctx, domain, fn) {
 
   var req = ctx.req
     , session = ctx.session
+    , waitingForCallback = false
     , callbackCount = 0
     , events;
 
@@ -94,6 +95,7 @@ Script.prototype.run = function (ctx, domain, fn) {
   if(domain) {
 
     events.on('addCallback', function() {
+      waitingForCallback = true;
       callbackCount++;
     });
 
@@ -140,7 +142,7 @@ Script.prototype.run = function (ctx, domain, fn) {
   }
   err = err || scriptContext._error;
   process.nextTick(function () {
-    if (callbackCount <= 0) {
+    if (!waitingForCallback && callbackCount <= 0) {
       done(err);
     }
   });
--- a/lib/script.js
+++ b/lib/script.js
@@ -14,13 +14,13 @@ function Script(src, path) {
 
 Script.prototype.runWithContext = function (context) {
   var functionArgs = Object.keys(context);
-  
+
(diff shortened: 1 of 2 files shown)


andreialecu:
Relevant code snippet for some sort of event:

```javascript
    var async = require("async");

    $addCallback();
    
    async.each(body.storeData, function (data, done) {
      doStuffOnData(data, function (err, result) {
         done(err, result);
      );
    }, function (err){
        if (err) return cancel(err);
        setResult({done: true});
        $finishCallback();
    });
```

ericfong:
Thanks for your contribution!
Landed as d8dbf3b0f3b0dfc6d0e0b18913c67f84814cafc1

Any good way to detect the functions which don't play well with dpd?
Use ```setTimeout``` and check ```callbackCount``` and show warning to dev ? But for each event will be too heavy?

andreialecu:
Deployd only attempts to wrap functions that are part of its script domain or script context.

Using `require()` bypasses that entirely, so everything async that is used through `require` needs this.

fotoflo:
in your snippet
```
    var async = require("async");

    $addCallback();

    async.each(body.storeData, function (data, done) {
      doStuffOnData(data, function (err, result) {
         done(err, result);
      );
    }, function (err){
        if (err) return cancel(err);
        setResult({done: true});
        $finishCallback();
    });
```

will return cancel be returned on error? how does $finishCallback() get reached?

andreialecu:
@fotoflo: `cancel` automatically finishes all callbacks and returns the response immediately as per https://github.com/deployd/deployd/pull/602

fotoflo:
Thanks

Rifazi:
@andreialecu you are literally my hero.

moorthi07:
This helps in the 'POST' event also. 
Use case:
Post event:

var createzipfile = require('..createzipfile');
  $addCallback();
try{

                    createzipfile(this,session,ctx,  (err,result) =>{
            if (err) {
              console.log('dispatch call failed', err);
                  cancel(" Error while generating app. Pl. try again. " + err);
         
          } 
         
              this.filepath= "{'zippath':'" + result +"'}";
          
               console.log('this.codemigration..vvvv.....',this.codemigration);
          
        $finishCallback();
          });
       
               }
               catch(err)
               {
                     console.log('err in dispatch------',err);
                   cancel(" Error while generating app. Pl. try again. " + err);
               }

PR #592, opened 2015-06-29
feature(collection): Add `previous` to AfterCommit event

andreialecu:
Also fix POST so that AfterCommit is called after successful database insertion rather than before it.

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -480,8 +480,13 @@ Collection.prototype.save = function (ctx, fn) {
 
   var domain = createDomain(item, errors);
 
+  domain.protectedKeys = [];
+
   domain.protect = function(property) {
-    delete domain.data[property];
+    if (domain.data.hasOwnProperty(property)) {
+      domain.protectedKeys.push(property);
+      delete domain.data[property];
+    }
   };
 
   domain.changed =  function (property) {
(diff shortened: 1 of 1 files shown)


andreialecu:
Usage example:

On AfterCommit:
```javascript
dpd.log.post({data: data, previous: previous, method: method});
```

lib/resources/collection/index.js
```diff
@@ -637,11 +649,18 @@ function createDomain(data, errors) {
...
   if (collection.shouldRunEvent(collection.events.AfterCommit, ctx)) {
-    collection.events.AfterCommit.run(ctx, {data: data, 'this': data, method: method}, function (err) { 
-      if (err) debug('AfterCommit errors in script: %j', err); 
+    data = _.clone(data);
+    if (protectedKeys && protectedKeys.length > 0 && previous) {
+      // add back whatever fields were protected, because they are removed from data
+      protectedKeys.forEach(function (key) {
+        data[key] = previous[key];
+      });
+    }
```
ericfong: Line 655-661 
```
    data = _.clone(data);
    if (protectedKeys && protectedKeys.length > 0 && previous) {
      // add back whatever fields were protected, because they are removed from data
      protectedKeys.forEach(function (key) {
        data[key] = previous[key];
      });
    }
```
only will be exec when called from line 563
```
collection.doAfterCommitEvent('PUT', ctx, item, prev, domain.protectedKeys);
```

Shall we move this part of code to there instead?

andreialecu: Not sure why that would be preferrable. The code is related to the AfterCommit event and I think it makes more sense to be inside that function rather than adding more code to the already long enough caller.

ericfong:
Thanks for your contribution!
Landed as fd8c9750922ebfeb35ab9b9c31ba9056c3687bf5

PR #593, opened 2015-06-29
feature(collection): Add Collection.extendDomain()

andreialecu:
Static function that allows adding values or functions to the domains of all collection instances.

Replaces https://github.com/deployd/deployd/pull/315

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -35,6 +35,7 @@ function Collection(name, options) {
 util.inherits(Collection, Resource);
 Collection.external = {};
 Collection.prototype.clientGeneration = true;
+Collection.domainAdditions = {};
 
 Collection.events  = ['Get', 'Validate', 'Post', 'Put', 'Delete', 'AfterCommit'];
 Collection.dashboard = {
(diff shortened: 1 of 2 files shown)


andreialecu:
You have access to `this.collection` and `this.domain` from inside any extensions defined using this method.

Usage example: 

app.js
```javascript
var Collection = require('deployd/lib/resources/collection');
var deployd = require('deployd');

Collection.extendDomain("getColName", function(){
  return this.collection.name;
});

Collection.extendDomain("hasChanges", function(){
  var domain = this.domain;
  if (_.find(_.keys(domain.data), function(key) {
    return domain.changed(key);
  })) {
    return true;
  }
  return false;
});

// configure database etc.
var server = deployd({
    port: process.env.PORT || 9000,
    env: process.env.ENV || 'development',
    db: {
        connectionString: process.env.MONGO_URI
    }
});

// start the server
server.listen();

// debug
server.on('listening', function () {
    console.log("Server is listening on port: " + (process.env.PORT || 9000));
});

// Deployd requires this
server.on('error', function (err) {
    console.error(err);
    process.nextTick(function () { // Give the server a chance to return an error
        process.exit();
    });
});
```

You can then use those defined functions in your event scripts, example:

On POST
```javascript
   emit(getColName() + ":post", this);
```

ericfong:
WOW. Great Feature!!!

andreialecu:
Yeah, it has a bit of dependency on #592. I can decouple if needed, but I also have another PR I need to send after these three which is related to AfterCommit as well.

So let's merge #592 first, that's easier.

ericfong:
Thanks for your contribution!
Landed as 450b8adae23770ff7899fe8dceef971cdcb539c9

PR #594, opened 2015-06-29
feature(internal-client): pass through headers and connection

andreialecu:
The internal-client did not pass through headers and connection data from the initial calling context, so things like the IP address and User Agent etc were lost with internal calls.

This properly passes through the context.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -12,13 +12,13 @@ var internalClient = require('./internal-client')
  * - **url** `String` The url of the request, stripped of the resource's base path
  * - **body** `Object` The body of the request, if the body is JSON or url encoded
  * - **query** `Object` The query of the request
- * 
+ *
  * @param {Resource} resource
  * @param {HttpRequest} req
  * @param {HttpResponse} res
  * @param {Server} server
  */
- 
+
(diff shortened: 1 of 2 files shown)


ericfong:
For me, it is good to pass ctx down. (actually, I think most event should not require http headers or connection information)

But, I think we should wait few days. So that the original writer can comment on it. (may do by intention)

andreialecu:
I use this in concert with #593, like this:

app.js
```javascript
Collection.extendDomain("getUserContext", function(ctx) {
  var userContext = {};
  if (ctx.req.connection) userContext.ipAddress = ctx.req.connection.remoteAddress;
  userContext.headers = ctx.req.headers;
  return userContext;
});
```

Then I have a `dpd-event` that posts to a different collection. In that collection, I have an aftercommit event that does something like:

```javascript
dpd.log.post({...., userContext: getUserContext(ctx) });
```

So, the IP address and browser header info is saved to my log colection.

Without this PR, ctx.headers and ctx.connection would be null.

ericfong:
Seems objection, let go

Landed as 9a1cefc0865412fedc14bf158b07ea595ca4a3e4

PR #596, opened 2015-07-03
feature(collection): add BeforeRequest event

andreialecu:
New event that allows inspecting a request prior to the database being touched, and before every other event is executed.

Allows canceling the request early (for example for authorization purposes), or modifying the query before it is executed by mongo depending on custom logic written in the event.

Gives access to `me`, `ctx`, `event` (one of: GET, PUT, POST, LOGIN, LOGOUT) variables in the event script. You can use all the `cancel()` related methods and end the request right there.

On PUT or POST, there are two additional variables available: `this` and `data` - containing the item being saved.

### DIFF
--- a/lib/resources/user-collection.js
+++ b/lib/resources/user-collection.js
@@ -59,8 +59,13 @@ UserCollection.prototype.handle = function (ctx) {
   }
 
   if(ctx.url === '/logout') {
-    if (ctx.res.cookies) ctx.res.cookies.set('sid', null, {overwrite: true});
-    ctx.session.remove(ctx.done);
+    var logoutDomain = { event: "LOGOUT" };
+    uc.addDomainAdditions(logoutDomain);
+    uc.doBeforeRequestEvent(ctx, logoutDomain, function(err) {
+      if (err) return ctx.done(err);
+      if (ctx.res.cookies) ctx.res.cookies.set('sid', null, {overwrite: true});
+      ctx.session.remove(ctx.done);
+    });
     return;
   }
 
@@ -126,7 +131,12 @@ UserCollection.prototype.handle = function (ctx) {
     break;
     case 'POST':
       if(ctx.url === '/login') {
-        this.handleLogin(ctx);
+        var loginDomain = { event: "LOGOUT" };
+        uc.addDomainAdditions(loginDomain);
+        uc.doBeforeRequestEvent(ctx, loginDomain, function(err) {
+          if (err) return ctx.done(err);
+          uc.handleLogin(ctx);
+        });
         break;
       }
       /* falls through */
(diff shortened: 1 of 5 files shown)


andreialecu:
Example of a hard limit of 20 records per GET query:

```javascript
if (!me) cancel("Not authorized", 401); // don't allow anyone in that is not authorized

switch (event) {
    case "GET":
        if (ctx.query && ctx.query.$limit) {
            ctx.query.$limit = ctx.query.$limit > 20 ? 20 : ctx.query.$limit; // max 20 results
        }
        break;
    case "PUT":
        // something for put
        break;
    case "POST":
        // something for post
        break;
}
```

andreialecu:
Also another example:

Say instead of deleting a record entirely, you just set `deleted: true` on it, in a field. This will make sure all GET requests return just non-deleted records:

```javascript
switch (event) {
    case "GET":
        ctx.query = ctx.query || {};
        ctx.query.deleted = {$ne:true};
        break;
} 
```

Previously, there was no good way to enforce this. If you did `cancelIf(this.deleted)` in the `GET` event and combined it with $limit, mongo would limit results to your $limit, but then deployd would filter deleted items out via the script, so you'd end up with an array of `$limit - number of deleted records` back. Which wasn't right.

Let's not merge this for a few more days, as I'm still testing it so I might make small changes. 

I will leave a note here when I think it's fully ready.

andreialecu:
Alright, @ericfong:

Been testing this for the past couple of weeks and it is stable on our end. If anything comes up, we can fix it later. It doesn't affect anything at all unless it is used.

Add ?w=1 to the link for easier code reviewing:
https://github.com/deployd/deployd/pull/596/files?w=1

lib/resources/user-collection.js
```diff
@@ -126,7 +132,12 @@ UserCollection.prototype.handle = function (ctx) {
     break;
     case 'POST':
       if(ctx.url === '/login') {
-        this.handleLogin(ctx);
+        var loginDomain = { event: "LOGOUT" };
```
ericfong:
```
var loginDomain = { event: "LOGIN" }; 
```
??

ericfong:
Great Thx! that is super useful for GET. 

Any guideline for using BeforeRequest with Validate in POST or PUT ?

lib/resources/user-collection.js
andreialecu: Oops.

andreialecu:
BeforeRequest does not replace Validate. It is still recommended to validate updates in the Validate event.

You can do some checks inside BeforeRequest that apply to all of PUT/POST/DELETE if you need to, mostly for authorization purposes or other special requirements your scenario may require.

ericfong:
Thanks for your contribution!
Landed as  4960d07fbade755470ee6d7ff29b9ccc7645c941

docnoe:
Awesome! 
I wrote the "dpd-beforeget" module to be able to cancel GET requests early. Your solution however is much more felxible, thank you!!

PR #598, opened 2015-07-05
fix(collection): object queries on booleans would be converted to false

andreialecu:
Before this fix, trying to do a query like {$exists: true}, or {$ne: true} on a boolean would replace the entire query with "false", leading to very hard to track bugs.


NicolasRitouet:
Nice job :+1:

NicolasRitouet:
Landed as 84a1d039833a023e22b98d5f4707f131d2f4f8d9
Thanks @andreialecu :trophy:

PR #602, opened 2015-07-14
feature(script): improve cancel()

andreialecu:
Allows passing an error object to `cancel(err)`, or an exception object returned by the deployd internal-client (eg. {message: "", status: 400}).

This allows easy chaining such as (in an event script):

```javascript
// call some other collection via internal-client
// both cases are handled by the new cancel, whether the code throws an Error object, or if .get returns an exception of the form {message, status}
dpd.somecol.get({...}).then(function() { error may get thrown here }.catch(cancel);
```

### DIFF
--- a/lib/internal-client.js
+++ b/lib/internal-client.js
@@ -123,7 +123,11 @@ exports.build = function(server, session, stack, ctx) {
                 callback(data, null);
               }
             } else {
-              callback(null, data);
+              try {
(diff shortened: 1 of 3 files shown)


ericfong:
Thanks for your contribution!
Landed as ea2bb0110a2aa55e9127a5547d943577eb79fb22

Issue #151, opened 2013-02-06
Deployd crash when testing PUT method with RESTConsole Google Chrome extension


iskugor:
Hi there. :)

I found (what seems to be) serious issue.

So ... you'll need to install Google Chrome extension RESTConsole: https://chrome.google.com/webstore/detail/rest-console/cokgbflfommojglbmbpenpphppikmonn

Setup Deployd collection, like "user" collection. Add some properties and data to it. And then try to update some property with RESTConsole:

* set request URI
* request parameters (properties you wish to update)

Click on "PUT".

You should get:

{
    "message": "Cannot read property 'id' of undefined",
    "status": 500
}

And Deployd should crash.

It seems that setting Content-Type header to "application/json" fixes this issue (Deployd does not crash), but another issue exists:

Could not parse invalid JSON

although, JSON seems to be valid.

Could you please check this issue? Seems pretty serious if Deployd can be crashed by just not sending proper header.

Thanks, Ivan.

ericfong:
I think this bug is fixed.

garfbargle:
$ dpd --version
0.8.4
$ node --version
v0.12.7
$ mongod --version
db version v3.0.2

Using postman chrome app and still seeing same exact issue

andreialecu:
@garfbargle I just tried reproducing this but couldn't. I get a `400: bad request` error from deployd but there is no crash.

ericfong:
no crash here too

garfbargle:
@ericfong @andreialecu here is how to reproduce:

1. $ dpd create my_crash_app
2. $ dpd -d
3. # add a user collection with GUI in browser
4. $ curl -X POST -H "Cache-Control: no-cache" -H "Postman-Token: ae4e53b3-18de-7a6b-e2d1-bd7b4b4009a1" -H "Content-Type: multipart/form-data; boundary=----WebKitFormBoundary7MA4YWxkTrZu0gW" -F "username=username" -F "password=password" 'http://localhost:2403/users'

NOTE: the curl request is what is generated by the postman app for chrome. I understand that it probably has some unnecessary information but this request specifically is causing my whole app to crash.

andreialecu:
@garfbargle Thanks, could reproduce it now. It only happens with user-collection. Fix at PR #603.

garfbargle:
@andreialecu That was fast! :) Thanks

PR #603, opened 2015-07-14
fix(user-collection): null check to prevent crash

andreialecu:
Add null check to prevent a crash with malformed requests under certain circumstances:
https://github.com/deployd/deployd/issues/151


ericfong:
Thanks for your contribution!
Landed as 36c4ec2df705278cdb7de16c774bd3b938b35a1d

PR #605, opened 2015-07-15
chore: update socket.io

andreialecu:
This permanently fixes an issue with installing on Windows, for which we had the version pinned to a specific commit previously.


ericfong:
Thanks for your contribution!
Landed as 36c4ec2df705278cdb7de16c774bd3b938b35a1d

PR #606, opened 2015-07-15
feature(http): cache OPTIONS request for 5 minutes

andreialecu:
This should reduce the number of OPTIONS requests for CORS calls to deployd. Set at five minutes, because Safari won't respect higher values anyway, and Chrome caps it at 10 minutes.


ericfong:
Great! can remove many OPTIONS

Thanks for your contribution!
Landed as 89cd1f023fddee118162d4efc055d8409fd9e899

PR #607, opened 2015-07-15
feature(http): option to allow dpd-ssh-key via CORS

andreialecu:
Required in certain circumstances (such as running unit tests across domains/ports)

Used like:

```javascript
var server = deployd({
    port: port,
    env: process.env.ENV || 'development',
    db: {
        connectionString: process.env.MONGO_URI
    },
    allowCorsRoot: true // allow root access (dpd-ssh-key) via cors
});
```


ericfong:
shall we use 'allowCorsSshKeyHeader' or 'allowCorsRootRequest' instead of 'allowCorsRoot'

andreialecu:
Renamed to `allowCorsRootRequests`, to go with [Root Requests](http://docs.deployd.com/docs/collections/reference/http.html#s-Root Requests)

ericfong:
Thanks for your contribution!
Landed as 2f795565036a20b731bfd0d0d3bdaa8fea3f5140

PR #608, opened 2015-07-17
feature(internal-client): allow access to underlying resource

andreialecu:
Allows access to resource internals from dpd internal client. 
Adds `getCollection()` method to store, which gives access to mongodb directly so you can run raw queries, for example from [dpd-event](https://github.com/deployd/dpd-event) scripts

Allows event script syntax such as:

```javascript
// callback
dpd.colname.getResource().store.getCollection(function(err, col) {
  // use col directly here, see mongodb documentation
});

// promise
dpd.colname.getResource().store.getCollection().then(function(col) {
  // use col directly here, see mongodb documentation
});
```

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -155,19 +155,32 @@ function getConnection(db) {
 function collection(store, fn) {
   var db = store._db;
 
-  getConnection(db).then(function (mdb) {
-    mdb.collection(store.namespace, function (err, collection) {
-      if(err || !collection) {
-        error(err || new Error('Unable to get ' + store.namespace + ' collection'));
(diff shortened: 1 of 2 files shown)


ericfong:
Thanks for your contribution!
Landed as d478b348d62be16fa2801aaff5f40d376fb9f193

Some thoughts about this commit:
1. dpd event should be fully flexible. So if you want to directly query DB, you can do so.
2. For events of collections, they should not do direct query. 
3. Sometimes, it is better to setup kue / other background queue
4. Use monk + co-monk is good to implement those background job

Issue #524, opened 2015-02-26
dpd fails after start in OSX


jsphstls:
$ dpd
    starting deployd v0.8.0...
    listening on port 2403
    type help for a list of commands
    dpd > Failed to start MongoDB (Make sure 'mongod' are in your $PATH or use dpd --mongod option. Ref: http://docs.deployd.com/docs/basics/cli.html)

As instructed I try:

    $ dpd --mongod
    starting deployd v0.8.0...
    child_process.js:1129
      var err = this._handle.spawn(options);
                         ^
    TypeError: Bad argument
        at TypeError (native)
        at ChildProcess.spawn (child_process.js:1129:26)
        at exports.spawn (child_process.js:988:9)
        at Object.exports.restart (/usr/local/lib/node_modules/deployd/lib/util/mongod.js:38:14)
        at start (/usr/local/lib/node_modules/deployd/bin/dpd:149:16)
        at Object.<anonymous> (/usr/local/lib/node_modules/deployd/bin/dpd:278:31)
        at Module._compile (module.js:460:26)
        at Object.Module._extensions..js (module.js:478:10)
        at Module.load (module.js:355:32)
        at Function.Module._load (module.js:310:12)
        at Function.Module.runMain (module.js:501:10)
        at startup (node.js:129:16)
        at node.js:814:3

I am able to run `$ mongod` without issue. I have a guess that the problem is.

MongoDB only works well on osx if you create /data/db as instructed in their install guide `mkdir -p /data/db`. But, Deployd instructs MongoDB to use the `/data` directory inside the app that I create with `dpd create` ? Everytime Mongo starts, it looks for the original `/data/db` while Deployd is working with the app `/data` ?

ericfong:
what is the output of ```which mongod``` ?

or please try  ```dpd --mongod `which mongod` ```

jsphstls:
$ which mongod
    /usr/local/bin/mongod

and then

    $  dpd --mongod `which mongod`
    starting deployd v0.8.0...
    listening on port 2403
    type help for a list of commands
    dpd > Failed to start MongoDB (Make sure 'mongod' are in your $PATH or use dpd --mongod option. Ref: http://docs.deployd.com/docs/basics/cli.html)

CWalig:
I had the same problem, I have to use sudo all the time.

Have you tried ?
```
sudo dpd
```

ericfong:
When you use sudo. Can you see "data" folder created in the project folder?

[...]

PR #616, opened 2015-07-23
fix(bin/dpd): prevent process quit when latestversion is not writable

### LINKED ISSUE #599: automatically quit version 0.8.4 on OSX
version 0.8.4
There is a problem that when I type `dpd` in the OSX terminal. It will automatically come out 
```
$ dpd
starting deployd v0.8.4...
listening on port 2403
type help for a list of commands
dpd > bye
```
I didn't enter Ctrl+c or other keys.

ericfong:
Fix #599 and #524

### DIFF
--- a/bin/dpd
+++ b/bin/dpd
@@ -298,7 +298,9 @@ function checkForUpdates() {
 
       if (json && json['dist-tags'] && json['dist-tags'].latest) {
         var latest = json['dist-tags'].latest;
-        fs.writeFile(latestversionFile, latest);
+        fs.writeFile(latestversionFile, latest, function(err){
+          // Need this callback function to prevent the process being killed, when latestversionFile is NOT writable
+        });
       }
     }
   });


bin/dpd
```diff
@@ -298,7 +298,9 @@ function checkForUpdates() {
 
       if (json && json['dist-tags'] && json['dist-tags'].latest) {
         var latest = json['dist-tags'].latest;
-        fs.writeFile(latestversionFile, latest);
+        fs.writeFile(latestversionFile, latest, function(err){
+          // Need this callback function to prevent the process being killed, when latestversionFile is NOT writable
```
andreialecu: Shall we console.log this as a warning?

ericfong: Some people use 
```
sudo npm install -g deployd
```
in mac or many be linux. 

And by default, /usr/local/lib/node_modules is not writable. 
Therefore, that is by default behaviour, I don't think we should have a warning for that.

Actually, we may complete remove this latestversionFile check. As people should check module updates by themselves.

Or, change the latestversionFile location to somewhere ~/.deployd/latestversion. But that will have other issues

PR #626, opened 2015-08-04
fix(internal-client): add null check for ctx

n321203:
Added check for ctx variable.

Ctx is undefined when using internal client from a custom module. This would cause a `TypeError: Cannot read property 'req' of undefined` on line 80. 

This patch fixes this by checking for ctx.


andreialecu:
Thanks for your contribution!
Landed as https://github.com/deployd/deployd/commit/d3ccf38ffd3c79afc8d6cb5c4a100ee0a2ba4472

PR #634, opened 2015-08-07
fixes #629: Property types in dashboard are sortable

### LINKED ISSUE #629: Dashboard type dropdown is sortable like the properties
### Description
In the properties panel of a collection, all datatype dropdowns are sortable. You can click-drag and rearrange the types, something causing them to disappear and empty properties to be added to the collection. (This also breaks the data panel of the collection).

### Examples:
http://imgur.com/a/iFGH3

Sortable datatype list items, that disappear on rearrangement
Sortable

Random junk property that appears after sorting
Funny Property

What the property looks like when you save and come back
Empty Property

A broken data panel if you let the empty property name persist
Broken Data Panel

docnoe:
Adds class "locked" to list items that should not be sortable.

### DIFF
--- a/lib/resources/collection/dashboard/properties.html
+++ b/lib/resources/collection/dashboard/properties.html
@@ -84,7 +84,7 @@
       <span class="caret"></span>
     </a>
     <ul class="dropdown-menu" data-bind="foreach: $root.propertyTypes">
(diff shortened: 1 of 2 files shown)


ericfong:
Thanks for your contribution!
Landed as ab7cec6302a901c6f73612b610ac6b559ad1081f

PR #637, opened 2015-08-17
Expose require() to scripts

n321203:


ericfong:
Would you mind add a little test in the script to test the require is working (may be with relative path and loading a module)

I test can help we ensure this feature work in future. Thanks a lot.

andreialecu:
Added in #757

PR #639, opened 2015-08-19
using require() without the command line

rgolea:
So I've been having this issue where I couldn't require modules from a project whenever I was running my own server. I added a couple of features where you can do `var module = require('name_of_the_module')` inside the app (basically I added dpd-event-extension to work natively with deployd) and I added an extra where you can do `var myCustomModule = requireLib('lib/test.js')` so you can add custom modules.

Hope you like it and if you have another thing in mind just feel free to reject this PR

### DIFF
--- a/lib/script.js
+++ b/lib/script.js
@@ -2,6 +2,7 @@ var _ = require('underscore')._
   , EventEmitter = require('events').EventEmitter
   , fs = require('fs')
   , bluebird = require('bluebird')
+  , br = require("base-require")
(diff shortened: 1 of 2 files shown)


ericfong:
Is this related to #637 ?

rgolea:
Have you used the #637 pr?

ericfong:
Sorry, forget what is going on. Do we still need this ?

rgolea:
I think not since you merged the other PR.

PR #642, opened 2015-08-22
fix: return error message if posted username is empty

docnoe:


ericfong:
please update the commit message and PR title to 
https://github.com/deployd/deployd/blob/master/CONTRIBUTING.md#type

So that following maintainers can understand your commits easier

docnoe:
Sorry, still a newbie on contributing to public repos.  Hope it's better now.

Issue #640, opened 2015-08-19
Bug /users/me crash


guyskk:
**First I create some users, then login with an user(saved cookie), then delete that user, at last, GET /users/me, the server crash**

Because I request  GET /users/me when webpage loaded, the server crash when I open my webpage.
As long as the cookie exist, the server will crash.

Here is the error messaes:
```
GET /users/me TypeError: Cannot read property 'username' of undefined
    at UserCollection.getUserAndPasswordHash (C:\Users\kk\AppData\Roaming\npm\no
de_modules\deployd\lib\resources\user-collection.js:297:46)
    at C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\lib\resources\user-c
ollection.js:115:29
    at C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\lib\db.js:385:11
    at handleCallback (C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\node
_modules\mongodb\lib\utils.js:95:12)
    at C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\node_modules\mongodb
\lib\collection.js:1122:5
    at handleCallback (C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\node
_modules\mongodb\lib\utils.js:95:12)
    at C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\node_modules\mongodb
\lib\cursor.js:597:5
    at handleCallback (C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\node
_modules\mongodb\node_modules\mongodb-core\lib\cursor.js:244:5)
    at setCursorNotified (C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\n
ode_modules\mongodb\node_modules\mongodb-core\lib\cursor.js:558:3)
    at C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\node_modules\mongodb
\node_modules\mongodb-core\lib\cursor.js:661:16
    at queryCallback (C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\node_
modules\mongodb\node_modules\mongodb-core\lib\cursor.js:321:5)
    at bound (domain.js:254:14)
    at runBound (domain.js:267:12)
    at Callbacks.emit (C:\Users\kk\AppData\Roaming\npm\node_modules\deployd\node
_modules\mongodb\node_modules\mongodb-core\lib\topologies\server.js:95:3)
[...]

PR #645, opened 2015-08-28
Prevent server from crashing if username doesn't exist

rgolea:
I had some problems where if the user had been deleted and the session token was still sent, the server would crash. This fixes that problem.


NicolasRitouet:
@rgolea thanks for your PR.
Could you add a test case with this?

rgolea:
I will as soon as possible. I took some vacations.

NicolasRitouet:
Enjoy then :sunny:

docnoe:
I would prefer PR https://github.com/deployd/deployd/pull/647, which fixes the same problem, but in a much simpler way.

ref to #640

NicolasRitouet:
#647 has been merged, it should solve the issue.

PR #647, opened 2015-09-13
prevent crash on invalid user for session

nicholasareed:
If a user was deleted manually, a GET for the logged-in user would crash the process. This simple fix puts the hash generation syntax in line with other requests in this file (`null` if !user).


docnoe:
fixes #640

NicolasRitouet:
Hi @nicholasareed 
Thanks for your PR!

Can you add whitespaces (`var userHash = user ? uc.getUserAndPasswordHash(user) : null;`)
and a test case?

NicolasRitouet:
Landed as 4183b7c904889b70b703ac7324f40bb1cc457ca0

PR #648, opened 2015-09-16
Fix/iss646: can't rename or delete new, empty collection

### LINKED ISSUE #646: Cannot rename or delete collection
After creating a collection, you cannot rename it and it displays an error, and if you try deleting the collection after that, it refuses to delete and shuts down deployd. below is the screen shot: Hope someone would look into this soon thanx.
screenshot from 2015-09-06 10 15 05

docnoe:
After creating a new collection it is not possible to rename or delete it if there was never any data added to it.
This is happening because the mongodb collection does not get created before the first entity is inserted.

To fix this, any error returned by "store.rename()" is checked. If it indicates that no mongodb collection by the given name exists it will not pass the error on to the dashboard, which will in return be able to properly rename or delete the resources folder and config.

fixes #646

### DIFF
--- a/lib/resources/collection/dashboard/properties.html
+++ b/lib/resources/collection/dashboard/properties.html
@@ -84,7 +84,7 @@
       <span class="caret"></span>
     </a>
     <ul class="dropdown-menu" data-bind="foreach: $root.propertyTypes">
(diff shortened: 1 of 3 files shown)


NicolasRitouet:
Hi @docnoe 
Thanks for your contribution ! :+1: 
Landed as 49f7696006461c2f4c07ec8bd6b04c277b405669

PR #649, opened 2015-09-16
Fix/iss640: Server crashes when cached cookie exits for deleted user and "GET users/me" is performed

docnoe:
See https://github.com/deployd/deployd/issues/640


docnoe:
Different (and more elegant) solution with same result is already in @nicholasareed PR: https://github.com/deployd/deployd/pull/647/files
Didn't notice until now so I will close my PR.

PR #672, opened 2015-10-28
fix(session): use update for updating sessions, not remove and insert

iKristjan:
Session update is currently implemented with remove and insert. During the update when the session is removed for a moment and then inserted again, a concurrent query can fail finding the session and result as an anonymous query.

It took a while to find the source of the problem, because it’s difficult to test it. It can happen only after every 10 seconds when the "lastActive" timestamp is updated and when you have another query at the same time, which resulted unauthorized response in my case.

This PR fixes the issue so that new sessions are inserted and old sessions are updated as they should, which allows concurrent queries to find the active session and authorize the user.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -312,22 +312,22 @@ Session.prototype.set = function(object) {
 Session.prototype.save = function(fn) {
   var session = this
     , data = _.clone(this.data)
-    , needsQueueBind = false;
-  var sid = null ;
+    , anonymous = false
+    , sid = null;
+  
   if (data.anonymous) {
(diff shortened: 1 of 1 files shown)


andreialecu:
@iKristjan did you accidentally close this?

iKristjan:
@andreialecu, yes I did. Unfortunately I let a critical bug in.

On row 347 "data.id" should have been "sid".
Using missing "data.id" caused updating all sessions.

I'll create a new pull request with updated and tested code.

iKristjan:
New fixed PR of this is #682.

PR #690, opened 2015-12-09
Cleanup: Remove grunt and grunt deps

NicolasRitouet:
@ericfong @andreialecu 
What do you guys think?
It's only used for jshint (I can add a npm task for that).


andreialecu:
I'm good with this. `jshint` is already being ran when running the tests.

ericfong:
Hi, I am good with this too.

NicolasRitouet:
Landed as 372b29c57b853ae675458ba2da8b21d6e3607d59

PR #691, opened 2015-12-09
Added Query Key for Mobile compatibility

yusufiga:
With this commit you can send advanced queries with “mongoq” key for
mobile url compatibility also this changes has backward compatibility.

example usage
/todos?mongoq={"category": "pets"}

### DIFF
--- a/lib/util/http.js
+++ b/lib/util/http.js
@@ -141,12 +141,12 @@ var parseBody = exports.parseBody = function(req, res, mime, callback) {
   });
 };
 
+
 var parseQuery = exports.parseQuery = function(url) {
   var q = url.substr(url.indexOf('?') + 1);
-
   if(q) q = decodeURIComponent(q);
-
-  if(q[0] === '{' && q[q.length - 1] === '}') {
+  if((q.startsWith('mongoq={') || q[0] ==='{') && q[q.length - 1] === '}') {
+    q=q.startsWith('mongoq={') ? q.substring(7, q.length): q;
     return JSON.parse(q);
   } else {
     return parseNumbersInObject(qs.parse(parseUrl(url).query));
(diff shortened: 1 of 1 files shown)


andreialecu:
There are some small issues with the implementation here. Deployd already supports a `_method` query string override, see [the docs here](http://docs.deployd.com/docs/collections/reference/http.html#s-HTTP method override). With your implementation, the order is dependent so the `mongoq` always has to come first, otherwise it is not recognized (if both _method and mongoq are present).

Also, the name `mongoq` ties this to mongo, and if we need to change dbs in the future it makes things awkward, so I would suggest `_query` or `_jsonquery`.

Could you also explain exactly why you needed this? What does this solve?

andreialecu:
Here's a simpler implementation of the above:

```javascript
var parseQuery = exports.parseQuery = function(url) {
  var q = url.substr(url.indexOf('?') + 1);

  if(q) q = decodeURIComponent(q);

  if(q[0] === '{' && q[q.length - 1] === '}') {
    return JSON.parse(q);
  } else {
    var parsedQuery = qs.parse(parseUrl(url).query);
    if (parsedQuery._jsonquery) {
      return JSON.parse(parsedQuery._jsonquery);
    }

    return parseNumbersInObject(parsedQuery);
  }
};
```

yusufiga:
I'm trying to use deployd api with Android. And some times I need to convert url to [URI object] (http://developer.android.com/intl/ru/reference/java/net/URI.html) like this
```
URI uri = URI.create("http://localhost?todos{\"category\": \"pets\"}");
```
but when try, I'm getting the following error because of this advanced mongo query in url {"category": "pets"}. So I try to add this "mongoq" key for fixing this error
```
Illegal character in query at index 22: http://localhost?todos{"category": "pets"}.
```

Thank you for feedback and implementation example. I will try to edit my codes based on your suggestion.

andreialecu:
Landed as https://github.com/deployd/deployd/commit/138b207415b976e0fe256d5536c2cd207b3c3aeb
Thanks for your contribution!

PR #693, opened 2015-12-10
fix: add class locked to password

kaala:
missing css property for password field


NicolasRitouet:
Landed as 3ab653373402022b4a8525652f046f4a42c14ad1

PR #696, opened 2015-12-20
Updated Loader to support sub modules

saleehk:
This will let us use multiple modules in same package 
just use in index.js 

``` javascript
 var module1 = require("module1"),
     module2 = require("module2");
 module.exports=[module1,module2];
```


andreialecu:
If you still need this, please comment and we'll discuss it.

PR #697, opened 2015-12-20
fix(session): memory leak 

andreialecu:
Sessions were not being removed from memory after being inactive. They were being retained forever.

Also, socket.io connections were not being properly removed from memory when they were closed.


NicolasRitouet:
Landed as a5e572a2504bc7f8230565be574d2107883fe1b1
Thanks @andreialecu !

PR #701, opened 2015-12-29
fix(server): disable resource type loader error

joanrieu:
The deployd server cannot run if any package listed in `package.json` cannot be `require()`d, which happens for some packages which are not meant to ever be `require()`d like `babel-runtime`. Any error due to a missing resource type will be reported when used, so it is not necessary to force-quit here.

I'm not completely happy with disabling error checks but the check makes the server completely unusable in this case.

### DIFF
--- a/lib/type-loader.js
+++ b/lib/type-loader.js
@@ -29,11 +29,13 @@ module.exports = function loadTypes(basepath, fn) {
             types[customResource.name] = customResource;
           }
         } catch(e) {
+          /*
           console.error();
           console.error("Error loading module node_modules/" + file);
           console.error(e.stack || e);
           if(process.send) process.send({moduleError: e || true});
           process.exit(1);
+          */
         }
 
         if(remaining === 0) {


andreialecu:
I think if you use a package.json file and put `babel-runtime` in your `devDependencies` instead of `dependencies`, it should work without this fix

joanrieu:
That's true for `babel-runtime` which is a compile-time dependency, but I bet there are other packages out there which will have the same problem and where this workaround won't be possible.

Maybe keep the code but display warnings without calling `exit`?

andreialecu:
It's unlikely any such packages should ever be in dependencies instead of devDependencies. 

Any *normal* packages should be able to be used through `require`, so they shouldn't have this problem.

Ideally this whole *load all packages* functionality should be deprecated, along with the CLI, and instead be able to specify exactly which packages are supposed to be deployd extensions through configuration. We talked with the maintainers about deprecating the CLI before, just didn't get around to doing it yet.

joanrieu:
I'm especially thinking about CSS-only packages distributed through npm.
I can only agree that explicite calls to `require()` (or though a config file) sound better anyway.

rgolea:
PR #718 fixes this in a more elegant way without having to disable catching errors. @joanrieu you need to have error handling in case something goes really wrong.

joanrieu:
I agree, disabling error checks is never a good thing, as stated in the PR. This was merely a stopgap until something better was implemented. The flag inside `package.json` is a better idea!

Issue #254, opened 2013-08-22
On put, if you protect() every field, will wipe out entire collection


brianhiss:
I was adding protect() to a put method, to secure a collection.  I started by adding all the fields and then backing out a few fields I wanted to change.  I noticed when I had all the fields in put.js listed with protect(), it wiped out every row of data in that collection.

I have been testing it several times with all different field types (array, string, number, etc.) and seems to be consistent behavior.

I am going to dig for a solution, but wanted to give a heads up if someone else experienced this or already has a solution.

Check out: [Protect in the Deployd Docs](http://docs.deployd.com/docs/collections/adding-logic.md#s-On Put-3789)

sigbjornt:
It does not look like this problem is resolved (using 0.8.4).
May I suggest to reopen the issue?


Small usecase where the object lost all values except from the id:
```
id="";

dpd.mycollection.post({Name:'Test', theArray:[{'data':'testdata'}]}, function(result, error) {
id=result.id
});

dpd.mycollection.put(id,{Name:"updatedName", theArray:[{'data':'updated-testdata'}]});
```

The collection has 3 fields: id (string), Name (string) and theArray (array).  The onPut event handler protects 2 fields
```
protect('theArray');
protect('Name')
```

PR #708, opened 2016-01-12
correct extreme uses of hide() and protect()

EvilDrW:
this addresses the problem of #254 where, if you hide or protect all fields of an object during a PUT operation, all the data of the record will be deleted.

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -568,7 +568,7 @@ Collection.prototype.save = function (ctx, fn) {
 
         delete item.id;
         if ((_.keys(prev).length > 0) && (_.keys(item).length === 0)) {
-          return done(null, {id: id});
+          return done(null, {});
         }
       	
         store.update({id: query.id}, item, function (err) {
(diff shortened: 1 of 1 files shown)


lib/resources/collection/index.js
```diff
@@ -567,6 +567,10 @@ Collection.prototype.save = function (ctx, fn) {
         }
 
         delete item.id;
+        if ((_.keys(prev).length > 0) && (_.keys(item).length === 0)) {
+          return done(null, {id: id});
```
andreialecu: I think this should return an empty object `{}` instead of the `id`. Maybe the id is intended to be hidden.

EvilDrW:
that is a good point.  changed to match.

andreialecu:
Would you mind rebasing this on master so I can merge it?

Thanks!

EvilDrW:
i'm confused... does something still need to be done?  i don't see anywhere that it says it needs to be rebased

NicolasRitouet:
Merged, thanks for your contribution 👍

PR #710, opened 2016-01-19
Added utf-8 encoding

GuilhermeMoreira:
Added utf-8 encoding to support pt-br and other languages accents on json response.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -74,7 +74,7 @@ Context.prototype.end = function() {
 
 Context.prototype.done = function(err, res) {
   var body = res
-    , type = 'application/json';
+    , type = 'application/json; charset=utf-8';
 
   // default response
   var status = this.res.statusCode = this.res.statusCode || 200;


rgolea:
Quite nice! Can you do it so you can extend that? Some people might need another charset

NicolasRitouet:
We already had the exact same PR for that a few months ago:
https://github.com/deployd/deployd/pull/318

GuilhermeMoreira:
Sorry I didn't see this other same PR.

Well, as pointed out then, I was having a problem with correctly displaying pt-br accents in the json response in Chrome, at least. Adding charset=utf-8 fixed it.

rgolea:
@NicolasRitouet would an optional parameter be better?

andreialecu:
This only seems to be an issue with the Chrome developer console. In actual XHR responses things come up fine. Probably a chrome bug: https://code.google.com/p/chromium/issues/detail?id=438464

rgolea:
So, the reason I asked was that these are the unsupported languages for utf-8. Now I don't speak any of these languages but might it be useful to allow people to support these languages? Would be any problem in the current deployd structure? This is now just a curiosity and not really something related to this PR.
http://www.unicode.org/standard/unsupported.html

juodumas:
JSON is always unicode, no need to add `charset=utf-8`. Moreover, JSON can be in `UTF-16` and `UTF-32`, not only in UTF-8: http://security.stackexchange.com/questions/82913/application-json-content-type-character-set. I believe this pull request can be closed.

As for unsupported languages, I think that unicode supports most and if there is a need, people using unsupported languages will go through the Unicode Consortium. The page you linked to states:
> Even though Unicode doesn't yet support these scripts, some of the languages which employ them are also commonly written with other scripts that are supported by Unicode.

NicolasRitouet:
@juodumas thanks for your comment, I agree that we should close this PR.

PR #712, opened 2016-02-01
MongoDB 3.2 is more strict on data type of .skip .limit - ensure that it is an integer

hjanuschka:
mongo 3.2 is more strict on $skip and $limit

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -597,8 +597,8 @@ function stripOptions(query) {
   var options = {};
   if(!query) return options;
   // performance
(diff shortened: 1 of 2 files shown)


rgolea:
Can you write a test for this?

hjanuschka:
there we go :octopus: 

 * added simple test

andreialecu:
Landed as https://github.com/deployd/deployd/commit/86e6f5f6a06d1c338974735d237fc9d8f33c8ec9

Thanks @hjanuschka! :+1:

PR #716, opened 2016-02-11
Supporting JavaScript Date type in Deployed

mobi-life:
Adds the ability to define Collection property fields with Date type ,
and makes it possible to do date/time operations in queries like sorting
base on date fields or searching base on date ranges . also makes it
possible to pass Date object to "post" or "put" methods , for example :
dpd.theCollection.post({articleDate:new Date()})


rgolea:
Can you please do a test for this as well?

mobi-life:
I'm not sure but I will try...
I'm new to nodejs and don't know the node unit testing tools well , but I did manual tests against these changes and worked well on my production server .

rgolea:
Yes, but for an open-source framework, it's way better to have a unit test...

andreialecu:
Superseded by #759

Issue #380, opened 2014-07-23
`dpd -d` fails when scanning node_modules directory


aahutsal:
```
archer@captain ~/tmp $ mkdir yo-dpd
archer@captain ~/tmp $ cd yo-dpd/
archer@captain ~/tmp/yo-dpd $ yo deployd
archer@captain ~/tmp/yo-dpd $ npm install grunt-autoprefixer
archer@captain ~/tmp/yo-dpd $ dpd -d
starting deployd v0.7.0...

Error loading module node_modules/grunt-autoprefixer
Error: Cannot find module '/home/archer/tmp/yo-dpd/node_modules/grunt-autoprefixer'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at /home/archer/dev/deployd/lib/type-loader.js:39:27
    at process._tickDomainCallback (node.js:463:13)
archer@captain ~/tmp/yo-dpd $ 
```

PR #731, opened 2016-05-17
`dpd -d` fails when scanning node_modules directory

eelcocramer:
I have a similar issue as raised in #380. It seems that this issue was closed automatically when the issuer fixed the issue in his own fork of `deployd` but there was never a PR to fix this upstream.

This PR has that fix but I applied it on the current master.


andreialecu:
Included in #737

Issue #727, opened 2016-04-13
npm domain marked as soon to be depracted


hjanuschka:
Stability: 0 - Deprecated : https://nodejs.org/api/domain.html

any idea wich module might step in?

PR #737, opened 2016-06-30
refactor(loaders): remove deprecated node domain calls

andreialecu:
Fixes https://github.com/deployd/deployd/issues/727

Removing the catch all error handling shouldn't matter much here, if there are any problems here they happen in the initialization step, so if anything is wrong we can let it crash anyway.


andreialecu:
I'll go ahead and merge this one since it's only a small change.

PR #740, opened 2016-07-16
fix(package): pin mongodb to 2.1.18 (upstream bug)

andreialecu:
The 2.1.21 version of the mongo driver has a race condition bug with authentication sometimes not being sent properly before a query is sent, resulting in deployd crashing.

It usually occurs under high traffic conditions.


hjanuschka:
@andreialecu thank you - for your work - this was the reason why we had to disable auth on mongo-side 👎

andreialecu:
Upstream bug ref: https://jira.mongodb.org/browse/NODE-722

PR #742, opened 2016-07-18
feat(http): allow handling custom mime types

andreialecu:
Example:

```javascript
var server = deployd({
  ...,
  mimeParser: function(req, res, mime, next) {
     if (mime === "my/mime") {
         req.body = myCustomParser(req);
         req.on('end', function() {
           next();
         }
     } else {
         next();
     }
  }
```


houmark:
Running this in production for months and it works just fine so +1 from me to get this PR in.

hjanuschka:
@andreialecu can you tell me a use-case - looks interesting but i don't get it 🎱 ?

houmark:
@hjanuschka I can explain a bit, since I was involved in having this feature developed.

Say you need to receive a POST request with an XML payload or another mime type. That's currently not possible.

Then in your app.js or whatever runs your server you can have the above posted code with this in addition:

Replace the my/mime above line with:

`if (mime === 'text/xml') {`

And add this function:

```
function xmlParser(req) {
  deploydUtil.getBody(req, function(b) {
    req.rawBody = b;
    req.body = b;
  });
}
```

In the above case you'd need to require deploy util lib:

`var deploydUtil = require('deployd/lib/util/http');`

This allows for the XML payload to pass through in example to a dpd-event script which can then handle the payload in its raw condition.

In our case we use this to receive POST webhooks/requests from a payment provider (Worldpay).

Hope this clears it up a bit :)

hjanuschka:
@houmark  thx for taking the time!
ok i got it. (already having a few ideas, on how to connect legacy api's to deployed using a custom mime parser.)

houmark:
@hjanuschka Yes, these few short lines opens up deployd for doing all that kind of stuff :) If you have time to test and give a +1 after testing I'm sure that would be appreciated so we can get this one merged.

PR #747, opened 2016-08-11
username case-insensitive check

Mortgy:
username case-insensitive check ( avoid duplicated users with similar usernames and different case sensitive letters )

i think it's a must to have this check
i.e
1 - user enter his username with first letter uppercase
2- user try to create another user with same username but all lowercase or different letter cases
3 - account will get created, usually users are not used to differentiate between lowercase / uppercase letters in usernames, only in passwords they care most.

hope it's useful pull request

### DIFF
--- a/lib/resources/user-collection.js
+++ b/lib/resources/user-collection.js
@@ -155,7 +155,8 @@ UserCollection.prototype.handle = function (ctx) {
       if(ctx.query.id || ctx.body.id) {
         this.save(ctx, done);
       } else {
-        this.store.first({username: ctx.body.username}, function (err, u) {
+        var usernameRegex = new RegExp(["^", ctx.body.username, "$"].join(""), "i");
+        this.store.first({username: usernameRegex}, function (err, u) {
           if(u) return ctx.done({errors: {username: 'is already in use'}});
           uc.save(ctx, done);
         });


lib/resources/user-collection.js
```diff
@@ -155,7 +155,8 @@ UserCollection.prototype.handle = function (ctx) {
       if(ctx.query.id || ctx.body.id) {
         this.save(ctx, done);
       } else {
-        this.store.first({username: ctx.body.username}, function (err, u) {
+        var usernameRegex = new RegExp(["^", ctx.body.username, "$"].join(""), "i");
```
andreialecu: There is a problem here if ctx.body.username contains regex control characters. The input would need to be escaped for this to work properly, otherwise the username can't contain a dot, $, ^. [ and a bunch of other characters. 

Using a regex here also hurts performance, since an index can't be used.

I think the solution is more delicate than this, an internal `username_lowercase` field should be populated/used instead.

Mortgy: so this way u need to force username at account creation / logging in to be lowercased so u can just keep comparing text at account creation ?

Mortgy:
so this way u need to force username at account creation / logging in to be lowercased so u can just keep comparing text at account creation ?
i can still create accounts that includes @, dot

rgolea:
I think the best way for this should be to make deployd ensure on table creation that a [unique index](https://docs.mongodb.com/v3.2/core/index-unique/) exists on the username field. Could we do that instead of querying?

I'm doing that on an app and it doesn't crash so maybe just adapt the errors a bit? What do you guys think?

weareutag (approved):

PR #753, opened 2016-08-31
Bluebird 3.4.4

houmark:


andreialecu:
Actually I don't see why this is needed since the semver range is ^3.4.3, so it should pick 3.4.4 up.

houmark:
True, but I guess deployd should often update the minimum requirement when the package has improvements included. I'm not sure an update always updates to the latest patch version.

On another note. Node 0.12 did not pass tests so there may be something changed in Bluebird that is not compatible with Node 0.12.

Maybe it's time for deployd to go v1.0 so semver can also be in full play and v1.0 could also drop support for Node 0.12. What do you think?

NicolasRitouet:
tests are unstable, they randomly fail 👎 
I restarted the tests and it passes now.
I think before going 1.0, we need to improve a few stuff to make deployd more reliable/secure on production.

But if the community is pushing it, why not.

houmark:
A lot of platforms uses this in production for a long time.

But v1.0 does not mean it's stable, it means v1.0 :) Nothing else.

The best way to adjust and tweak features is to be above v1.0, so semver can handle breaking changes, because semver does not consider anything below v1.0 stable in any way and that means it will not be able to warn about breaking changes from version to version. If releases follow semver fully and version is 1.0+ tools like `npm-check` will warn when a breaking change is added do deployd.

I can recommend seeing [this video](https://www.youtube.com/watch?v=tc2UgG5L7WM) and start using [semver release](https://github.com/semantic-release/semantic-release) for all releases as it takes care of bumping the version according to the commits since last release and their type, please it creates a nice changelog.md fully automatic.

NicolasRitouet:
@houmark thanks for your input.

According to semver (which we are already using), you should release 1.0 when your software is used on production.
Some people are using it on production, but we (the maintainer's team) don't think that it's ready for production, especially for the security aspect.
If the crowd decides that we are stable enough to release a 1.0, I'm fine with it since we already follow the semver ideas (https://github.com/deployd/deployd/blob/master/HISTORY.md).
If it helps, I can create a ticket about a 1.0 release to ignite the discussion.
Since it's a different discussion, I'll close this ticket.

PR #754, opened 2016-09-02
Add response headers setting

p3nt4:
I find it useful being able to modify response headers on the server level. 
I tried an implementation of it,.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ var Server = require('./lib/server')
 module.exports = function (config) {
   var server = new Server(config);
   upgrade(server);
+  server.options.responseHeaders=config.responseHeaders;
   return server;
 };
 
--- a/index.js
+++ b/index.js
@@ -8,7 +8,7 @@ var Server = require('./lib/server')
 module.exports = function (config) {
   var server = new Server(config);
   upgrade(server);
-  server.options.responseHeaders=config.responseHeaders;
+  server.options.responseHeaders = config.responseHeaders;
   return server;
 };
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -105,7 +105,7 @@ Server.prototype.handleRequest = function handleRequest (req, res) {
     if(server.options.hasOwnProperty("responseHeaders")) {
 	for (var header in server.options.responseHeaders) {
     	     if (server.options.responseHeaders.hasOwnProperty(header)) {
-        	res.setHeader(header, server.options.responseHeaders.header);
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -8,6 +8,7 @@ var Server = require('./lib/server')
 module.exports = function (config) {
   var server = new Server(config);
   upgrade(server);
+  server.options.responseHeaders=config.responseHeaders;
```
houmark: Maybe add a space on each side of the `=`?

lib/server.js
```diff
@@ -102,6 +102,13 @@ Server.prototype.handleRequest = function handleRequest (req, res) {
 
   // add utilites to req and res
   setupReqRes(server.options, req, res, function(err, next) {
+    if(server.options.hasOwnProperty("responseHeaders")){
```
houmark: Missing space before last `{`.

houmark:
Can you provide a use case on how to add headers so I can test it out when I have time?

p3nt4:
`// staging.js
var deployd = require('deployd');

var server = deployd({
  port: 8081,
  env: 'dev',
  db: {
    host: '127.0.0.1',
    port: 27017,
    name: '',
    responseHeaders:{
	 Pragma: 'no-cache'
     },
    credentials: {
      username: '',
      password: ''
    }
  }
});

server.listen();

server.on('listening', function() {
  console.log("Server is listening");
});

server.on('error', function(err) {
  console.error(err);
  process.nextTick(function() { // Give the server a chance to return an error
    process.exit();
  });
});
`

lib/server.js
```diff
@@ -102,6 +102,13 @@ Server.prototype.handleRequest = function handleRequest (req, res) {
 
   // add utilites to req and res
   setupReqRes(server.options, req, res, function(err, next) {
+    if(server.options.hasOwnProperty("responseHeaders")) {
+	for (var header in server.options.responseHeaders) {
+    	     if (server.options.responseHeaders.hasOwnProperty(header)) {
+        	res.setHeader(header, server.options.responseHeaders.header);
```
andreialecu: Shouldn't this be `res.setHeader(header, server.options.responseHeaders[header]);` ?

andreialecu:
This is a bit specific. 

You could achieve the same outcome by using a plugin like [dpd-router-middleware](https://github.com/ezra-obiwale/dpd-router-middleware) and using `ctx.res.setHeader(...);` from the event script.

Going to close this in favor of the above approach.

PR #757, opened 2016-09-22
Added require to the script context

gabrielperales:
I have added require to the script context and I wrote a test to check if require is a function inside the context of the script. I think I don't have to test other thing because require is well known function already tested by the node team.

I know that more than one year ago there was another pull request doing the same but without adding tests, and now there it is outdated and there will be merge conflicts, so I have thought it will be better to make a new pull request than fixing the other one.

There are a lot of outdated pull request than there should be closed.

Also, I decided to make this pull request because I saw that this was on the milestone https://github.com/deployd/deployd/milestone/6 .

I hope this can be helpful.

### DIFF
--- a/lib/script.js
+++ b/lib/script.js
@@ -82,6 +82,7 @@ Script.prototype.run = function (ctx, domain, fn) {
       return (scriptContext.me && scriptContext.me.id === id) || false;
     },
     console: console,
+    require: require,
     query: ctx.query,
     internal: req && req.internal,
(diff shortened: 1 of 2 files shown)


andreialecu:
Hey, could you please update this to latest master? I'll merge it after.

Thanks.

gabrielperales:
@andreialecu I just have rebased master. I think it should be ok.

PR #776, opened 2016-10-26
refactor: remove dpd cli

andreialecu:
This moves the cli to its own package which can be installed globally separately.

It allows deployd to be installed locally rather than globally in a project, which we should recommend going forward.


andreialecu:
The cli is on npm as `dpd-cli` and should be installed globally by users via `npm install dpd-cli -g`

andreialecu:
Repository is at https://github.com/deployd/deployd-cli

NicolasRitouet:
This PR has been merged on the `next` branch.
We'll test it there until it's ready to go on `master`.

PR #777, opened 2016-10-26
fix: server.attach() did not enable socket clustering

Michael-Hamburger:
PR to enable redis.socket-io usage when using attach

### DIFF
--- a/lib/attach.js
+++ b/lib/attach.js
@@ -63,15 +63,14 @@ function attach(httpServer, options) {
   // back all memory stores with a db
   server.db = db.create(options.db);
 
-  var socketServer = io.listen(server, _.extend({
-    'log level': 0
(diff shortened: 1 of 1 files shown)


houmark:
I'm not exactly sure this is needed.

We are running socket.io with redis using attach and that works just fine.

This is how we do it:

```javascript
var server = require('https').createServer(options, app);
var io = require('socket.io')(server, { transports: ['websocket', 'polling'] });
var redis = require('socket.io-redis');
io.adapter(redis({ host: 'localhost', port: 6379 }));
```

and then:

```javascript
deployd.attach(server, {
  socketIo: io,
  env: ENV,
  db: {
    //
  }
});
```

Would like to hear what's the benefit of adding this to deployd core?

Michael-Hamburger:
Server.js does it like that. Attach.js should do the same.

PR #782, opened 2016-11-03
fix(collection): protect and hide should not modify `this`

andreialecu:
## Description
Calling `protect('property')` in an event script would delete the property from the `this` object, so `this.property` would immediately become undefined after a protect call.
Same with hide.

For example:

```
On Put:

// this.owner has a value
protect(this.owner);

// this.owner was undefined here (bug):
if (this.owner == me.id) { ... }
```

Additionally, calls to `protect()` in PUT would supress those properties from being returned in the response JSON, which was a bug. The proper way is to use `hide()` for such purposes.

There is a potentially breaking change for users of the library that were relying on the broken behavior of `protect()` to hide fields from being returned via `PUT`. However, that was never documented, so it should be an infrequent use case.

## Motivation and Context

With this change, if after protecting a property, some other code attempts to update it, a `TypeError` error will be thrown. But this should be expected.

## How Has This Been Tested?
The PR includes several new tests.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/lib/resources/collection/index.js
+++ b/lib/resources/collection/index.js
@@ -492,29 +492,7 @@ Collection.prototype.save = function (ctx, fn) {
     fn(errors || err, item);
   }
 
-  var domain = collection.createDomain(item, errors);
-
-  domain.protectedKeys = [];
-
-  domain.protect = function(property) {
-    if (domain.data.hasOwnProperty(property)) {
-      domain.protectedKeys.push(property);
-      delete domain.data[property];
-    }
-  };
-
-  domain.changed =  function (property) {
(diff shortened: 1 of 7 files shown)


andreialecu:
The PR also adds the ability to unprotect and unhide fields via either `protect('field', false)`/`unprotect('field')` or `hide('field', false)`/`unhide('field')`

PR #801, opened 2017-02-15
Update to 2017

Phineas:


NicolasRitouet:
@Phineas can you also update it on other places?

PR #820, opened 2017-07-27
Chore: Release 1.0.0 🎉

NicolasRitouet:
# Checklist
- [x] Remove cli from core
- [x] Remove dashboard from core
- [x] Remove clientlib from core
- [x] Release deployd-cli
- [x] Release dpd-dashboard
- [x] Release dpd-clientlib
- [x] Write changelog
- [x] Update package.json version

@deployd/developers anything else I forgot before releasing this 1.0.0?


rgolea:
@NicolasRitouet Why is the deployd-cli on v.2.0.0? Now it's fine but I think we will have the same problem as the Angular team when they had the Angular Router on version 3 and the rest of the packages on version 2. People were posting issues because they were using with v.2. of Angular a version 2 of the router which had bugs on it. Maybe we can keep the same version?

andreialecu:
I also think there should be more in the changelog about how to use the new cli and dashboard packages. Unless there's documentation elsewhere that I didn't see.

NicolasRitouet:
@rgolea we had to use 2.0.0 for `deployd-cli` because of the previous project using this name.
I understand your concern. The best solution would be to switch `deployd` to 2.0.0 also, but that would be weird. Let's keep that in mind and do it later.
@andreialecu the cli is the same as before, we already have documentation for this, do you think we need something more detailed?
The dashboard package indeed needs more details, but at the moment, since the cli also adds the dashboard, it's transparent for the user. I would also focus on that after the release. What do you think?
Thanks for your feedback folks 👍

rgolea:
@NicolasRitouet maybe we could ask the community? If the community understands there is no relation between the cli version and the deployd version, then there is no need to bump them to the same version.
Also, I don't think it should be a problem in bumping deployd to v.2. Right now it's the perfect moment to do that because of possible breaking changes. Doing it later is more awkward than bumping to v.2 right now. If we bump it later, deployd will jump from v1 to v3 (at least).
Anyways, it not really a big thing. We can release it as it is.

NicolasRitouet:
@andreialecu @rgolea I added a short migration guide. Anything you'd like to add/improve before releasing?

PR #821, opened 2017-09-16
Chore: remove useless dependencies from dashboard

genu:
## Description

`filled` package depends on `"mime": ">= 1.2.6"`, but latest version of mime has breaking changes, as per the note on their repo: `Version 2 is a breaking change from 1.x`. `filled` didn't make the necessary updates, so deployed is broken because of it. Since deployd doesn't depend on latest version of `filled`, its safe to force the latest `1.x` version of mime as well.
## How Has This Been Tested?
Running `dpd -d` no longer throws an error

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)


NicolasRitouet:
Thanks for your proposal.
`filed`  is used by the dashboard, that we extracted from the core.
We should actually remove it from `package.json` and add `mime` dependency from [dpd-dashboard](https://github.com/deployd/dpd-dashboard)

NicolasRitouet:
Thx for your PR, before merging this one and releasing 1.0.0, I'll release a fix on 0.8.

PR #831, opened 2017-11-02
fix(tests): run tests on random mongo port and update for new mocha/npm

andreialecu:
Minor changes to get tests running reliably again


andreialecu:
Not sure why this fails on Node 9 on Travis, I tried it locally and it passes.

PR #851, opened 2017-12-15
fix: don't overwrite http headers with defaults

andreialecu:
## Description
This allows calling `ctx.res.setHeader('Content-Type', sometype)` from script code and ensuring deployd doesn't overwrite it.

## Motivation and Context
I needed to set the content type to 'text/javascript' for a particular response and it would get overwritten by a default.

## How Has This Been Tested?
Added tests

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -75,6 +75,10 @@ Context.prototype.end = function() {
 Context.prototype.done = function(err, res) {
   var body = res
     , type = 'application/json';
+  
+  // make sure the response hasn't already been sent
+  // might happen if someone calls ctx.res.end() inside an event script
+  if (this.res.finished) return;
 
   // default response
   var status = this.res.statusCode = this.res.statusCode || 200;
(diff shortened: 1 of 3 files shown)


moorthi07:
Is this is the reason for this,

> Error: Can't set headers after they are sent.

andreialecu:
@moorthi07 no, that warning will appear if you use `ctx.res.end()` by yourself, since deployd does it too, later down the request pipeline.

However, I can fix it in this same PR. I'll add a commit for that too.

moorthi07:
@andreialecu good point.
So, in custom event module (dpd-event) , I tried,
 ctx.res.setHeader('Content-Type', 'application/octet-stream');
              ctx.res.write(fs.readFileSync(distPath));
This works , but still get the error
Error: Can't set headers after they are sent.

PR #870, opened 2018-09-25
fix: To string on an object is showing [Object object] in the error page

christianguevara:
Will check if the object sent as part of cancel or done is an error, if it's an error object the .toString() in the HTML error template will print the right message, if the error sent is an object it will print [Object object] to avoid it is needed to get the err.message property to make the .toString() in the node-doh template to actually show the right message.

## Description
When a cancel('message', 200), the done() method transform this call to an object containing {message: 'message', statusCode: 200}, this object is sent to render in the HTML page of node-doh lib, where it's called object.toString() which causes  [Object object] to appear as part of the HTML and not the word 'message' or any string sent.

## Motivation and Context
This fix is required to actually show an error explanation to the user.

## How Has This Been Tested?
Actually there's no test for this since have no clue where to add it and also the test doesn't run fully even before the change. :/

## Screenshots (if appropriate):
Image showing the error:
cancel

Both cancel calls have the same result. 
## Types of changes
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.


andreialecu:
I fixed the other, unrelated problems with the tests, but it looks like this PR introduces some additional failures.

christianguevara:
Will comeback to this, but at the moment can't make tests run at all.
     MongoError: Use of SCRAM-SHA-256 requires undigested passwords
     MongoError: User 'foo@dpd-faux-remote' not found

Seems related to MongoDB version.

andreialecu:
You might be able to use docker for the tests. See https://github.com/deployd/deployd/pull/869#issuecomment-418321703

andreialecu:
The problem is when using mongodb 4.x with the 2.0 mongodb native driver. It's only broken because tests create a db user and the 2.0 mongo driver can't create users on mongodb 4.0.

Looks like they're not taking PRs for 2.x any more.
https://github.com/mongodb/node-mongodb-native/pull/1852

Everything else should work.

